Issue #73, opened 2026-04-19
feat: multi-worktree heartbeats PR-2 — discovery + watcher + CWD + concurrency


ryaneggz:
Implements PR-2 of .claude/specs/multi-worktree-heartbeats-spec.md.

Stacked on #71 (PR-1). Adds git-worktree-based discovery, N file watchers, cwd on spawn, HEARTBEAT_MAX_CONCURRENT semaphore.

ryaneggz:
Completed in PR #74 (merged 2026-04-20).

PR #99, opened 2026-04-22
FROM fix/98-entrypoint-slack-link TO development

### LINKED ISSUE #98: fix: entrypoint silently skips Slack bot build/link
## Problem

`install/entrypoint.sh:76-84` is supposed to build and `pnpm link --global` the vendored Slack bot (`packages/slack/`) so the `mom` CLI is on PATH inside the sandbox. In practice it never runs:

1. **Wrong path.** The block guards on \`[ -f \"\$HOME/harness/packages/slack/package.json\" ]\`, but entrypoint runs as root, so \`\$HOME=/root\`. The file lives at \`/home/sandbox/harness/...\` and the condition is always false — the block is skipped entirely.
2. **Swallowed errors.** \`pnpm install/build/link\` all redirect stderr to \`/dev/null\`, so even on the rare occasion the guard passes, any failure is invisible.
3. **Wrong user.** The commands would run as root and write to root's pnpm store — the \`sandbox\` user wouldn't see the linked binary on their PATH.

## Impact

\`mom\` (the Slack bot CLI) is absent from fresh sandboxes. \`install/onboard.sh\` then hits its \`command -v mom\` check (line 161), prints \`mom CLI not found\`, and Mom is never started — the Slack integration silently does nothing even though tokens are valid and \`.env\` persisted.

Observed in \`oh-local\` today: \`which mom\` returns empty; \`~/.claude/.onboarded\` shows \`slack: skipped\`; \`/usr/local/share/pnpm\` doesn't exist.

## Fix

- Use the absolute path \`/home/sandbox/harness/packages/slack\` in the guard.
- Run the pnpm commands via \`gosu sandbox\` so they use the sandbox user's pnpm store.
- Stop redirecting stderr; log clear success/failure messages from the entrypoint.

ryaneggz:
## Summary

- **`install/entrypoint.sh`**: the Slack-bot build/link step ran as root where `$HOME=/root`, so its guard on `$HOME/harness/packages/slack/package.json` was always false. Silent `2>/dev/null` redirections hid the skip. Result: `mom` CLI absent from the sandbox PATH on every fresh container, and onboard's Slack step silently no-ops. Fixed by using the absolute sandbox path, running via `gosu sandbox`, and surfacing success/failure.
- **`packages/sandbox/src/__tests__/heartbeat-discovery.test.ts`**: husky exports `GIT_DIR`/`GIT_INDEX_FILE` when it runs the pre-commit hook; those leaked into the test's child `git` processes and retargeted them at the real repo, so `git worktree add -b agent/sdr-pallet` failed (that branch already exists from #73). Strip `GIT_*` from `process.env` in a `beforeAll`.

Closes #98

## Test plan

- [x] `pnpm run test` in `packages/sandbox` passes (260/260)
- [x] Same test suite passes with `GIT_DIR=…` exported (simulates husky)
- [ ] After rebuilding `oh-local` (`docker compose ... up -d --build`), `which mom` resolves inside the container and entrypoint log shows `[entrypoint] Slack bot linked (/...path-to-mom)`
- [ ] Fresh `openharness onboard` flow sees `mom` on PATH and starts the bot

### DIFF
--- a/install/entrypoint.sh
+++ b/install/entrypoint.sh
@@ -73,14 +73,17 @@ if [ -f "$STARTUP" ]; then
   gosu sandbox bash "$STARTUP" 2>&1 | sed 's/^/  /' || true
 fi
 
-# Build and link Slack bot from bind-mount (replaces /opt/slack image copy)
-if [ -f "$HOME/harness/packages/slack/package.json" ]; then
-  echo "Building and linking Slack bot package..."
-  cd "$HOME/harness/packages/slack"
-  pnpm install 2>/dev/null
-  pnpm run build 2>/dev/null
-  pnpm link --global 2>/dev/null
-  cd "$HOME/harness"
+# Build and link Slack bot from bind-mount (replaces /opt/slack image copy).
+# Entrypoint runs as root, so $HOME is /root — use the sandbox user's absolute
+# path and run pnpm as that user so the global link lands on their PATH.
+SLACK_PKG="/home/sandbox/harness/packages/slack"
+if [ -f "$SLACK_PKG/package.json" ]; then
+  echo "[entrypoint] Building and linking Slack bot package..."
+  if gosu sandbox bash -c "cd '$SLACK_PKG' && pnpm install && pnpm run build && pnpm link --global"; then
+    echo "[entrypoint] Slack bot linked ($(gosu sandbox bash -lc 'command -v mom || echo missing'))"
+  else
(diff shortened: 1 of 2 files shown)


ryaneggz:
Superseded by #102 which bundles these commits with #101's settings.json default fix into a single merge. The entrypoint fix was cherry-picked; the test hygiene commit was re-authored to replace a contamination slip (the original `d230632` was committed under `test@example.com` due to a local git config left behind by a test run).

PR #522, opened 2026-06-24
fix: reinstall pnpm deps on manifest drift

### LINKED ISSUE #521: fix: refresh sandbox installs when manifests change
## Finding

Sandbox boot dependency setup skips installation whenever `node_modules` exists, even if `package.json`, `pnpm-lock.yaml`, or workspace package manifests changed since the last install. That can leave agents running with stale JavaScript dependencies until a human manually deletes modules or reruns installs.

## First-principles rationale

Open Harness depends on repeatable sandbox startup. A long-lived bind-mounted workspace and auth volumes make `node_modules` persistence useful, but the current existence-only check trades speed for correctness: dependency drift is invisible at boot. The fix should preserve the fast path when manifests are unchanged while forcing a deterministic reinstall when the root/workspace package manifests or lockfile change.

## Plan sketch

- Inspect `.devcontainer/entrypoint.sh` dependency setup and identify every root/workspace install fast path that keys only on an existing `node_modules` directory.
- Add a small manifest-fingerprint marker stored next to the relevant install output, covering root `package.json`, `pnpm-lock.yaml`, and workspace package manifests used by the sandbox.
- Reuse the existing fast path when the marker matches; run the install when the marker is absent or stale, then atomically refresh it.
- Emit clear boot log messages that distinguish `dependencies current` from `manifest drift detected; reinstalling`.
- Add or update script tests/probes for the fingerprint helper or boot logic so future changes cannot regress to existence-only checks.
- Document the behavior in the relevant sandbox/dependency README or inline comments if no dedicated docs section exists.

ryaneggz:
## Selection rationale

Research selection: queue had no actionable open `autopilot` ticket after local PR dedupe; `/harness-audit` ranked stale sandbox dependency installs as the top small, high-impact harness-infra finding. First-principles rationale: sandbox startup must be repeatable, and an existence-only `node_modules/` fast path can leave long-lived sandboxes with dependencies that no longer match manifests. Filed as #521.

---

## Summary

- add a manifest fingerprint marker for root/workspace pnpm dependency inputs
- rerun sandbox boot `pnpm install` when manifests or the lockfile drift despite existing `node_modules/`
- keep the fast path when the fingerprint is unchanged and log why install is or is not running

## Verification

- `bash -n .devcontainer/entrypoint.sh`
- `pnpm vitest run scripts/__tests__/entrypoint-pnpm-install.test.ts`
- `pnpm test:scripts -- scripts/__tests__/entrypoint-pnpm-install.test.ts` (runs the script test suite; 25 files / 285 tests passed)
- `bash .claude/skills/eval/run.sh` (58 probes; exit 0)

Closes #521


ryaneggz:
Closing after the PR-audit remediation pass.

**Why:** the change introduces a first-boot regression. The entrypoint forces a `pnpm install` on every fresh container boot because the manifest-fingerprint marker (`node_modules/.openharness-pnpm-manifests.sha256`) is always absent on a new container — even when a valid `node_modules` is baked into the image. The reinstall delays cron-session startup past the healthcheck window, so the **Sandbox Boot Guard fails** (container goes `unhealthy`: missing `cron-watchdog`/`cron-system`; ~7m43s vs development's ~28s, which skips the install because `node_modules` is present).

The drift-reinstall convenience (auto-reinstall when `package.json`/lockfile change) is minor — the entrypoint already reinstalls when `node_modules` is missing, and `pnpm install` can be run manually — and it isn't worth the boot-time fragility it added.

**If revived:** the clean fix is *adopt-existing* — when `node_modules` is present but the marker is absent, write the marker and skip the install (preserving the fast boot path and keeping drift-detection on later boots), or seed the marker at Docker build time after the baked install. Diagnosis preserved here for reference.

Issue #756, opened 2026-08-13
[EPIC #731] Close two host-root escape paths: operator config bind + docker socket


ryaneggz:
Two host-root paths out of the sandbox container are open in the running
deployment. Both are **deviations from what the repo already declares**, so
neither needs new architecture — the live container has to match its own
configuration.

Filed under [EPIC #731](https://github.com/mifunedev/openharness/issues/731).
Durable owner for the credential half is
[#734](https://github.com/mifunedev/openharness/issues/734).

## Summary

| # | Escape | Why it exists today | What retires it |
|---|---|---|---|
| 1 | Operator config directory bind-mounted read-write into the container | Keep secrets out of the repo | A container-private named volume — **already what compose declares** — plus #734 |
| 2 | Host `/run/docker.sock` bind-mounted read-write | The only way an in-container agent gets Docker | Sysbox's own in-container `dockerd` (RFC §9) |

Either path gives an in-container agent effective host root. The docker-socket
overlay says so in its own header:

> Mounting the host Docker socket gives the sandboxed agent effectively host
> root (it can start a privileged container that mounts the host filesystem),
> so enable this ONLY when the agent must drive Docker.

— `.devcontainer/docker-compose.docker-sock.yml`

## Escape 1 — operator config directory bind

**Evidence** (collected 2026-08-12 from inside container `34263ba23a57`):

- The container's mount table shows the operator's home config directory
  bind-mounted read-write onto the sandbox user's config directory. Container
  inspect reports `Type=bind`, `RW=true`, with the Docker Desktop WSL shim
  source path `/run/desktop/mnt/host/wsl/docker-desktop-bind-mounts/...`.
- Every **other** credential directory is a container-private named volume:
  `oh-sbx-local_claude-auth`, `_codex-auth`, `_pi-auth`, `_ssh-config`,
  `_herdr-data`. The config directory is the only one that reaches back to the
  host.

**This contradicts the checked-in compose.** `.devcontainer/docker-compose.yml`
declares the config mount as a plain named volume (`config-dir:`, line 87) with
no `driver_opts` and no bind. That volume exists but is not attached. Ruled out
as explanations:

- Compose used only the two repo files
  (`docker-compose.yml`, `docker-compose.docker-sock.yml`); the socket overlay
  adds *only* the socket.
- `.devcontainer/devcontainer.json` has no `mounts` key.
- `config-dir:` landed 2026-07-21 (`dd58de9d`, #659), **before** this container
[...]

Issue #731, opened 2026-08-10
[EPIC] Decouple Operator from execution environments


ryaneggz:
## Summary

OpenHarness **is the Operator**. The architectural goal is not to move OpenHarness onto the host or to make Sysbox itself the architecture. The goal is to separate the **Operator/brain** from the **development execution environment/hands** behind a stable runtime-neutral contract.

The default standalone shape should preserve one of OpenHarness's strongest properties: the Operator itself remains reproducible and containerized. A lightweight host-side `oh` launcher owns topology/lifecycle; the Operator runs in its normal container; a separate execution environment provides the repository, shell, builds, tests, Docker, and project services.

New information materially changes the preferred execution-runtime evaluation: **Docker Sandboxes use dedicated microVMs with a private Docker daemon under the hood** and are designed specifically for agentic development workloads. This makes Docker Sandboxes a first-class candidate for the standalone execution environment and means Sysbox should no longer be hard-wired as the target architecture.

The architecture is:

> **Operator outside a disposable full-development execution environment.**

Docker Sandbox, Sysbox, SSH/VM, Firecracker, and future managed sandboxes are implementations of that boundary.

## Current state

Today the primary sandbox combines:

```text
OpenHarness sandbox
├── OpenHarness / Operator
├── Claude / Codex / Pi / other providers
├── skills / agents / hooks / cron
├── provider auth + tool state
├── project repository
├── build/test tooling
└── optional host Docker socket access
```

This is convenient and reproducible, but the Operator and the machine it operates are the same thing. That creates several constraints:

- the provider process assumes project files/tools are local;
- execution-runtime changes become changes to the Operator environment;
- Docker capability may require exposing the host Docker socket;
- execution lifecycle and provider/auth lifecycle are unnecessarily coupled;
- remote/microVM execution requires special-case workflows rather than an execution contract.

## Product decision: preserve the containerized Operator

For standalone OpenHarness, **do not make host-native Operator the primary architecture**.

Moving Claude/Codex/Pi/OpenHarness directly onto arbitrary user hosts would reintroduce host dependency management for Node, provider CLIs, Python/tooling, PATH, permissions, process supervision, and upgrades. The current container provides valuable determinism and provider-auth persistence through named volumes.

Preferred standalone topology:

```text
                              HOST
                               │
                    lightweight `oh` launcher
                    topology + lifecycle only
                               │
                  ┌────────────┴────────────┐
                  │                         │
                  ▼                         ▼
[...]

PR #764, opened 2026-08-13
FROM task/762-health-check-host-side TO development

### LINKED ISSUE #762: task: rehome /health-check host-side now that the sandbox has no Docker socket
The host Docker socket is gone from the sandbox (#756). `/health-check` was the
one real in-container casualty and is now partly inoperable where it runs.

## What breaks

`/health-check` triages host memory, disk, CPU and Docker usage, ranks reclaim
levers, and prunes build cache. Everything that reaches Docker now fails:

```
docker ps
failed to connect to the docker API at unix:///var/run/docker.sock; check if the
path is correct and if the daemon is running: dial unix /var/run/docker.sock:
connect: no such file or directory
```

The memory, swap, disk and CPU checks still work in-container, but they measure
the container's view, not the host's — which was already the weaker half of the
skill's value.

## Options

1. **Rehome host-side.** Make it an operator-run skill that executes on the host,
   where both the Docker data and the real memory/disk figures live. Truest to
   what the skill is for.
2. **Document host-only and degrade cleanly.** Keep it invocable in-container but
   have it detect the absent socket and say so, instead of erroring per-command.
   Cheapest, and honest.

Option 2 is worth doing regardless, because an agent that runs the skill today
gets a wall of connection errors rather than one clear statement.

## Acceptance criteria

- [ ] Running `/health-check` in a socket-less sandbox produces a single clear
      statement that Docker triage is host-side, not a series of failed calls.
- [ ] The Docker-dependent steps are either relocated host-side or explicitly
      marked host-only in the skill.
- [ ] The non-Docker checks state that they measure the container, not the host.
- [ ] `AGENTS.md`'s skills table entry matches the new scope.

Refs #756, #731.

ryaneggz:
Closes #762. Refs #756, #731.

## What was broken

#756 removed the host Docker socket from the sandbox as one of two host-root escape paths. The `docker` CLI is still installed, so `/health-check` did not fail *once* — it failed **nine times per invocation**, each with the same 200-character connection error: three calls in step 2, one verbose call, `docker stats` in step 5, and four in step 6. Observed in this worktree's container:

```
$ ls -la /var/run/docker.sock
ls: cannot access '/var/run/docker.sock': No such file or directory

$ docker ps
failed to connect to the docker API at unix:///var/run/docker.sock; check if the
path is correct and if the daemon is running: dial unix /var/run/docker.sock:
connect: no such file or directory

$ command -v docker
/usr/bin/docker
```

Meanwhile the memory/disk/CPU steps went on reporting the **container's** numbers under host framing ("Report-first resource triage for the sandbox host", "Docker lives on the root overlay here, so root df is the binding number"). That is the more dangerous half: a green container disk row reads like permission to start a multi-GB image build.

## The decision, and why — both options, not one

The issue named two options. **Both land**, because they address two different readers of the same skill, and this harness already defines both.

Option 1 is not a hypothetical host runtime. Root `AGENTS.md` — reached as `CLAUDE.md`, which is a symlink to it, so it is always loaded — defines the **orchestrator**: an agent that runs at the host project root, is explicitly granted `docker`/`docker compose` in § Permissions, and already carries `/health-check` in its own skills table. That is the true addressee of the Docker half. So the rehome is concrete: Docker triage moves from "commands the sandboxed agent runs" to "the procedure the orchestrator runs at the host project root."

What #756 severed is narrower and is the actual defect — an agent *inside* the sandbox can no longer reach host state, by design. Option 2's clean degrade is what that reader gets.

Doing only Option 2 would leave the skill claiming to triage a host it cannot see. Doing only Option 1 would leave the sandboxed reader facing the nine-error wall.

*A first draft of this rationale claimed no host-side agent runtime existed. A critic showed the repo's own always-loaded context contradicts that. The correction improved the design rather than just the wording — the relocated procedure now has a named addressee that already has the repo checked out. Recorded in [`critique.md`](.oh/tasks/health-check-host-side/critique.md).*

## What changed

**`scripts/scope-preflight.sh` (new)** — classifies scope and endpoint **once** and emits `SCOPE`, `DOCKER_CLI`, `DOCKER_ENDPOINT`, `DOCKER_TRIAGE`, `METRICS_SCOPE`, plus exactly one `HEALTH-CHECK SCOPE-NOTICE:` line naming why, what was skipped, who runs the relocated procedure, and where it lives.

The design decision that matters: **`available` requires a completed `docker version` round-trip, not `[ -S path ]`.** A socket file outlives an OOM-killed daemon, and a `chmod 000` socket passes every file test while being unopenable — a critic demonstrated both. Either would have been classified `available` and reproduced the identical wall of errors *behind a passing preflight*, in a shape no test covered.

Three states, not four. An `unverified` state was designed and dropped: nothing forced it to ever be *resolved*, so a compliant implementation could report it and stop, leaving a `tcp://` operator exactly where the old skill left them. Collapsing it makes resolution structural — no endpoint, no daemon contact at all; any endpoint, exactly one round-trip.

Endpoint resolution follows the real CLI (`DOCKER_HOST` → `docker context inspect` → default socket → rootless path), because Colima, OrbStack and rootless setups set a *context* rather than the env var. The `context` read is free: verified to return the endpoint with `rc=0` while no daemon is running.

The script always exits `0` — a classification step that exits non-zero is itself the misleading-failure signal being removed.

**`SKILL.md`** — step 0 branches three ways; the `KEY=VALUE` lines are explicitly *not* the report (this skill's own habit elsewhere is to `tee` output into the answer); steps 2, 5, 7 and the reclaim ladder carry host-only markers; a new **Host-side Docker triage** section specifies the round trip in both directions, including what the report concludes when nobody pastes anything back (`UNKNOWN`, never silently pending).

**The verdict now refuses what it cannot support.** Labelling scope was not enough — a label explains a number without withdrawing a conclusion. Under `host-only` with a build-shaped target the Disk row renders `⚪ N/A` instead of a RAG rating computed from container `df`, and the report states that no build-sizing verdict is possible from inside the container. Container-local questions ("will `npm install` OOM this container") keep a real rating, because there the container view *is* the correct view.

Also: the false "root overlay is the binding number" claim is corrected rather than relabelled, and a dangling `/docker-disk-cleanup` pointer — a skill that exists nowhere in the repo — is removed from the section being rewritten.

**Alignment** — `AGENTS.md`, `.oh/templates/AGENTS.md`, the frontmatter `description` **and** `TRIGGER` list (leaving "free up space" pointing at a skill that can no longer reclaim in-container is a discoverability regression), and `link-providers.sh` `required_execs` so boot validation catches a missing or non-executable preflight.

`/audit full --health-target` now **decides** rather than offering an either/or: a socket-less composition is container-scope evidence recorded `partial`, deliberately **not** `deferred` — `deferred` promises a rerun recovers the evidence, and an in-container rerun produces the identical gap.

## Acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | Socket-less run produces a single clear statement, not a series of failed calls | ✅ One notice, `DOCKER_TRIAGE=host-only`, **0** Docker calls issued — proven by a shim, not by grep |
| 2 | Docker steps relocated host-side or explicitly marked host-only | ✅ Both — host-only markers **and** the Host-side Docker triage procedure |
| 3 | Non-Docker checks state they measure the container | ✅ `METRICS_SCOPE` carried into step 1, the verdict header, and the log entry |
| 4 | `AGENTS.md` skills-table entry matches the new scope | ✅ Plus the template and the always-injected frontmatter |

## Verification

- **`health-check-socket-degrade.sh` (new)** — 9 assertions across 6 arms. It *executes* the preflight: host-only classification; a `docker` shim proving **zero daemon contact** on that branch; the `available` arm; the `unreachable` arm (a dead endpoint must never be called available); a real bound unix socket; an absent CLI; no terminal `unverified` state; and that `SKILL.md` actually wires step 0 — an unwired script is a no-op however correct it is.
- **Verified by rejection, with attribution.** 11 mutations, each shown to fire its **own named assertion** rather than merely to turn the probe red. Two defects were found in the *test method itself* before it was trusted: the shim logged `-H`'s value instead of the subcommand (so one assertion would have passed for the wrong reason), and the absent-CLI arm broke its own `PATH` and silently skipped.
- **Full suite: 103 probes, zero `REGRESSION`/`ERROR`.** The diff against the pre-change scoreboard is exactly one added row. The 3 `SKIPPED` are pre-existing.
- The three probes that constrain this rewrite stay green: `health-check-docker-stats` (the `docker stats` step survives, relocated not deleted), `memory-log-locked-append` (**both** its assertions — `AUDIT_RUN_ID` *and* the locked-append line; a critic caught that my regression table had named only one), and `audit-dispatcher-contract`.

Evidence doc: [`.oh/tasks/health-check-host-side/evidence.md`](.oh/tasks/health-check-host-side/evidence.md)

## Notes for the reviewer

- **On the `title-convention` audit flag:** the audit was right and my first reading was wrong. I sampled the five newest PRs, saw `type: description`, and concluded the classifier's `^FROM .+ TO .+$` test matched nothing. Over the last 30 PRs, 20 match it, and [`/git` SKILL.md:82](.oh/skills/git/SKILL.md) states it literally — `FROM <source-branch> TO <target-branch>`. I had conflated it with the *commit* format (`<type>: <description>`), which the commits here do follow. Retitled; the flag clears.
- Provider surfaces are symlinks (`readlink .claude/skills` → `../.oh/skills`, same for `.pi`/`.codex`), so one file edit serves every surface. No per-surface copies. The `AGENTS.md` edit was confirmed live through the `CLAUDE.md` symlink.
- `.oh/scripts/sandbox-healthcheck.sh` and its open roadmap row are a **different subsystem** — container liveness reported through Docker/Compose health status — untouched here, and stated as a non-goal so the shared name does not invite folding them together.
- Restoring the socket is explicitly not on the table; #756 is closed.


ryaneggz:
## Final `/audit pr` — correlation closed

`evidence.md` records an audit run against a head that predates the file itself. This comment carries the confirming run against the head that **includes** it, which closes that correlation without regress.

| Field | Value |
|---|---|
| Head audited | `c3552a22` (final) |
| Run id | `audit-20260814T031158Z-1955959` |
| Verdict | **`PR-AUDIT-PROMOTABLE`** |
| `promotable` / `evidenceComplete` | `true` / `true` |
| Flags | *(none)* |
| Log terminal state | `complete`, exit `0` |
| CI | 4/4 `pass` — Boot Path Lint 23s, Eval Probe Regression Gate 28s, Lint/Typecheck/Build/Test 35s, Validate sandbox compose 1m45s |
| Merge state | `isDraft=false`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` |

### What changed since the previous audit

`development` moved a fourth time (`c95caf01`, prompt-miner score clamp). This branch merged it clean and regenerated the scoreboard.

- **The merge changed no implementation byte.** `scope-preflight.sh`, `health-check-socket-degrade.sh` and `SKILL.md` are byte-identical at `c3552a22` to the commit the 11-mutation attribution harness ran against (`4802b3ea`). That result stays binding on the code this PR merges.
- **The scoreboard diff is timestamps alone.** `ran 106 probe(s)`, rc=0, zero `REGRESSION`/`ERROR`, 4 `SKIPPED` all pre-existing. Every changed row is a `last-run` field, and no probe changed status. Normalising the timestamp out of the diff leaves each line present on both sides.
- **`CHANGELOG.md` kept both entries.** The fourth merge added one line and removed none.

### Re-run gates

`link-providers.sh --check` rc=0 · `health-check-socket-degrade` PASS · `pnpm run typecheck` clean · `pnpm run build:harness` `dist/oh.js 79.7kb` · `pnpm test:scripts` 42 files / 590 tests passed.

One gate moved owner, stated rather than assumed: this container has no `shellcheck` binary, so the CI **Boot Path Lint** job owns that verification on the final head. It passed.

PR #826, opened 2026-08-25
fix: make the installer's answers land in the file that wins

ryaneggz:
## Why

`install.sh` wrote every non-secret answer to `.devcontainer/.env` while its own closing text (`install.sh:556`) named `harness.yaml` as the file whose "active keys override local env" — so the user's answers landed in the **losing** file. Nine settings were affected: sandbox name, timezone, git name/email, and the five `INSTALL_*` flags.

Compose tolerated this, because `docker-compose.sh:118-160` falls back to `.env` for every overlay decision. **The `oh` CLI does not** — it resolves `sandbox.name` (`lifecycle.ts:274`) and `install.*` (`harness.ts:149-152`) from `harness.yaml` only. Symptoms, in likelihood order:

1. `oh harness list` reports `enabled: false` for a harness `install.sh` baked into the image.
2. `oh shell` / `oh stop` / `oh logs` target the wrong container — masked for the standard `~/.openharness` install (both default to the string `openharness`), but it **bites fork-and-clone users**, whom `installation.md:22-40` explicitly tells to run `install.sh` from inside their own clone.
3. `.oh/templates/harness.yaml` had silently diverged from `harness.yaml.example`, so several keys had no home to be written into at all (`setHarnessKey` no-opped on absent keys).

## What changed

**Config plumbing**
- `install.sh` seeds `harness.yaml` *before* prompting, then writes the nine non-secret answers into it via a new `_yaml_set` that respects `harness-config.sh`'s awk grammar (section header at column 0, key at exactly 2-space indent). Per-key fallback to `.env` if a write does not take.
- `DOCKER_SOCKET` stays in `.env` as a **documented exception** — the VS Code "Reopen in Container" path loads `.devcontainer/docker-compose.yml` directly and cannot read `harness.yaml`. `lifecycle.ts:162-164` already made the same call.
- Collapsed the two disagreeing `harness.yaml` line editors into one section-scoped `setKeyInSection` in `lib/harness-yaml.ts`. The wizard's section-blind copy silently dropped answers for keys absent from the template.
- Reconciled `.oh/templates/harness.yaml` with `harness.yaml.example` — restored `sandbox.docker_socket`/`image`/`pull_policy`, `install.opencode`/`grok_build`/`deepagents`, `paths.worktrees`, and the `crons`/`autopilot`/`slack`/`compose` sections.
- Added `ssh.enabled`/`ssh.port` and `sandbox.docker_socket` to the `oh init` wizard (now 4 steps), each defaulting off, the socket prompt carrying the host-root warning. Deliberately stopped there rather than chasing full key coverage.

**Docs**
- README and quickstart lead with the two scripted doors (`install.sh` for a sandbox of your own; `oh init` to equip an existing project). The untested manual clone sequence moved into the collapsed section — and its **duplicate** at `README.md:232-241` is merged away.
- `installation.md:7`'s "no host CLI, no Node toolchain" claim scoped to the clone path; the same file documents the Node ≥ 20 CLI at `:207`.
- `lifecycle-commands.md` gained a "which door am I?" table above the existing `make` vs `oh` section.
- `install.sh`'s next-steps names the file that received the answers first, and offers `oh harness install <name>` instead of the `make destroy && make sandbox` rebuild that #821 made unnecessary.
- Corrected `harness.yaml.example:23`'s claim that `pull_policy` reaches the VS Code path. It does not — and neither does any other key in the file. Authoritative statement: `.oh/docs/deployment-prebuilt-image.md:96-99`.

## Reviewer note — this reverses a tested invariant

`compose-args.test.ts` had a test named *"installer keeps host defaults out of tracked harness.yaml"*, asserting `expect(text).not.toContain('ok "harness.yaml: sandbox.name')`. Someone deliberately pinned the behavior this PR reverses. The test now pins the new contract with the reasoning in a comment above it — **but if that invariant was load-bearing for a reason not recorded in the repo, this is the commit to challenge.**

## Verification

- **Tests: 808/809.** The one failure (`cron-runtime.test.ts > buildTmuxWrapper`) **pre-exists on `development`** — confirmed by stashing and re-running. Untouched by this PR.
- **Probes: 107 pass, 3 skipped, 1 regression.** `audit-run-root-contract.sh` also **pre-fails on `development`** (same check).
- Two new probes, both passing: `harness-yaml-schema-parity.sh` (the two schemas carry the same key set; `setKeyInSection` is the only writer) and `oh-init-headless-config.sh` (`oh init --yes` prompts zero times and seeds a byte-identical `harness.yaml`, protecting CI/cron provisioning from future wizard growth). The parity probe was negative-tested by deleting a key — it fails and names it.
- `_yaml_set` exercised end-to-end against the real `harness-config.sh` parser: all six keys round-trip; a value containing `#` is correctly refused and falls back to `.env`.
- CLI `tsc --noEmit` clean; `bash -n install.sh` clean.

## Still open — needs a human, not the repo

1. **A 2-minute empirical check would falsify the core premise.** Set `sandbox.name: probe-me` in `harness.yaml` *only*, "Reopen in Container", then `docker ps`. If it reports `probe-me`, the VS Code path *can* read `harness.yaml` and the `DOCKER_SOCKET` exception here is wrong. Everything else in this PR stands either way.
2. **README headline positioning.** This PR gives `install.sh` door A on the grounds that it satisfies the repo's own no-Node prerequisite promise and ends with a running container. The counter is real: `oh init` is the far better-tested path (778 lines of behavioral tests vs. four greps over help text).


ryaneggz:
Closing as **already shipped**, not as rejected. Every line of this PR is on `development` — a two-dot `git diff origin/development` against this branch (after merging `development` in) returns **empty**.

It arrived by an unintended route. This branch's commit `8a107922` was in the ancestry of `feat/msb-runner-docs`, so squash-merging [#827](https://github.com/mifunedev/openharness/pull/827) carried the installer fix in as a passenger; [#817](https://github.com/mifunedev/openharness/pull/817) brought the rest. That is also why this PR went `CONFLICTING` — it was conflicting with its own content.

Verified present on `development`:

| Item | Status |
|---|---|
| `_yaml_set` in `install.sh` (answers written to `harness.yaml`, parser-verified, with fallback) | present |
| `setKeyInSection` in `lib/harness-yaml.ts` | present |
| `harness-yaml-schema-parity.sh`, `oh-init-headless-config.sh` | present |
| `harness.yaml.example` `pull_policy` / VS Code correction | present |
| README + quickstart installer-first rewrite | present |
| All five CHANGELOG bullets | present |

**One thing did not get resolved and should not be lost:** this PR flagged that its core premise — whether the VS Code "Reopen in Container" path can read `harness.yaml` at all — was unverified, and called it a two-minute empirical check. The shipped `harness.yaml.example` now asserts that it *cannot*. That assertion is live in `development` on the strength of an unrun check. Worth a follow-up issue rather than leaving it implied by a closed PR.

PR #869, opened 2026-08-29
FROM task/868-remove-oh-context-tier-upstream TO development

### LINKED ISSUE #868: [TASK] Remove the .oh/context tier
## Description

Delete the `.oh/context/` tier in full and retire the machinery that existed only to serve it.

An advisor council reviewed whether `.oh/context/` earns its keep versus `AGENTS.md` alone. Operator decision: the directory goes, and nothing in it is promoted.

Findings that make the deletion clean rather than lossy:

- **There is no loader.** No `SessionStart` hook exists; `.claude/settings.json` carries only `PreToolUse`/`Stop`/`Notification` hooks. The tier loaded only because `AGENTS.md` said "At the start of every session, read: …" — voluntary compliance, guarded by a probe that grepped the instruction text, never the read. Deleting the read-list deletes the entire load path.
- **The always-on tier is 25,091 B (~6,272 tokens).** Removing `.oh/context/` leaves `AGENTS.md` at 9,873 B — a 61% cut, and `AGENTS.md` is auto-loaded mechanically via the root `CLAUDE.md` symlink, so what remains is guaranteed rather than hoped for.
- **`SOUL.md` (119 B), `TOOLS.md` (200 B), `USER.md` (333 B) have no consumers.** Every reference is machinery that exists to measure or protect them.
- **`REPO_MAP.md` (11,775 B, 47% of the tier) was never proven.** `CB-004` has scored PARTIAL since 2026-07-03 on "Δ +0.00 machinery-added"; the A/B manifest, the scorer, and `ablate.sh` all exist and were never run. Its guard `repo-map-contract.sh` is a Goodhart instrument — ~15 literal-prose `grep -Fq` assertions that pass green while the routing table sends readers to `.pi/skills/*` provider mirrors instead of canonical `.oh/skills/*`.
- **`IDENTITY.md` was `/retro`'s only write target.** Deleting it makes `/retro` strictly report-only, matching the CHANGELOG trajectory and non-negotiable #5: a lesson that matters graduates to a **probe**, not to prose.

## Context

Not consistent enough across harnesses to be standard practice — that is the operative reason. Nested `AGENTS.md` files are rejected for the same reason and none will be created.

Scope, in five parts:

1. Delete `.oh/context/` and its load path in `AGENTS.md`. Relocate only `directory-readme.md` → `.oh/skills/harness-context/references/`.
2. Retire the machinery: `ablate.sh`, `ablate-state-machine.sh`, `repo-map-contract.sh`, CB-004 and its scorer; rewrite `context-tier-size-budget.sh` as a single-file `AGENTS.md` budget; clean `protected-paths.txt` and the CI path filters.
3. Make `/retro` report-only.
4. Repoint every remaining `.oh/context/…` reference (~40 files), including the `oh init` payload manifest so initialized projects stop shipping a dead tier.
5. Fix the dangling links found along the way.

**Out of scope:** root `README.md`, `.oh/README.md`, and `docs/README.md` are not reconsidered — `.oh/README.md` is touched only to correct paths that would otherwise dangle. The missing `.oh/agents/advisor.md` (cited as canonical by 4+ skills) is a separate issue.

---

> **Git workflow**: see the [/git skill](../../.oh/skills/git/SKILL.md)

---

## Done When

- [ ] `.oh/context/` no longer exists; `grep -rn "\.oh/context"` across tracked sources returns nothing
- [ ] The always-on tier is `AGENTS.md` and nothing else (~9,873 B, down from 25,091 B)
- [ ] Every touched probe is PASS and **not SKIPPED**, and each rewritten probe was deliberately broken to confirm it goes red on its own assertion message
- [ ] `/retro` emits a report and writes no file
- [ ] `oh init` into a temp dir produces no `.oh/context/`
- [ ] CHANGELOG `### Removed` entry names each deletion with its evidence and states `REPO_MAP.md` was removed as *unproven*
- [ ] PR targets `development` unless the git workflow rule says otherwise

ryaneggz:
Closes #868

Deletes the `.oh/context/` always-on context tier in full and retires the machinery that existed only to serve it. **`AGENTS.md` (9,374 B) is now the only always-on context file**, auto-loaded mechanically via the `CLAUDE.md` symlink rather than by prose request.

## Why this is a clean deletion

- **There was no loader.** No `SessionStart` hook exists. The tier was read only because `AGENTS.md` said "At the start of every session, read: …" — voluntary compliance, guarded by a probe that grepped the *instruction text*, never the read.
- **`SOUL.md`, `TOOLS.md`, `USER.md` had no consumers** — every reference was machinery measuring or protecting them.
- **`REPO_MAP.md` is removed as *unproven*, not disproven.** `CB-004`, its `repo-orientation` manifest, and its scorer landed 2026-07-03; the paired A/B workload was **never run once** in the ~2 months they stood, which is why the row held at `Δ +0.00 machinery-added`. Its guard `repo-map-contract.sh` was itself a Goodhart instrument — ~15 literal-prose assertions passing green while the routing table it guarded pointed readers at `.pi/skills/*` provider mirrors.

`directory-readme.md` is **kept and relocated** to `.oh/skills/harness-context/references/`, now carrying the routing rule it owns: descriptions stay in a directory `README.md`; directives move to the owning skill; only the root orchestrator contract lives in `AGENTS.md`; no nested `AGENTS.md` files.

## Two consequential decisions

**1. The ablation subsystem is retired, not just `ablate.sh` deleted.** `ablate.sh` is a *sourced library*, not a standalone script — `.oh/skills/eval/run.sh:35` sources it unconditionally under `set -euo pipefail`, with six consumers total. Deleting the file as-is would have taken down the entire probe suite. Its allowlist accepted **only** the five session-start context files, so removing the tier leaves it no legal target; `--ablate` is therefore stripped from `/eval` and `/audit context` and every source site removed coherently.

**2. `/retro` becomes strictly report-only.** `IDENTITY.md` was its only write target. A graduated lesson is now nominated as a candidate probe under `.oh/evals/probes/` rather than appended to identity prose — a lesson that matters graduates to a probe, not to prose. `CB-003` and the `DS-020` dataset, which scored that promotion cycle, retire with it.

## Verification

| # | Check | Result |
|---|---|---|
| 1 | `bash .oh/skills/eval/run.sh` | **92 probes, 89 PASS, 3 SKIPPED, 0 REGRESSION.** Baseline was 94/91/3. The 94→92 delta is exactly the two deleted probes (`ablate-state-machine`, `repo-map-contract`). |
| 1b | No probe silently became SKIPPED | ✅ Same 3 as baseline — `debugmcp-availability`, `next-dev-prod`, `registry-portability` — all `unchanged`, all environment conditions, not fixture loss. |
| 2 | Red-team each rewritten probe | ✅ See table below — every one failed on **its own assertion message**. |
| 3 | `git grep '\.oh/context'` | ✅ Clean. Surviving hits are all deliberate (see below). |
| 3b | Bare `context/` sweep (no `.oh/` prefix) | ✅ Clean. The plan's own grep missed these; sweeping for them caught 4 further dangling refs. |
| 4 | Always-on tier size | **9,374 B** (`AGENTS.md` alone), guarded by a new 9,500 B ratchet. |
| 5 | `bash .oh/evals/probes/protected-paths-resolve.sh` | ✅ PASS — all 31 entries resolve. |
| 6 | `bash .oh/scripts/link-providers.sh --check` | ✅ exit 0 — `.pi/.claude/.codex/.prime skills -> .oh/skills`; `CLAUDE.md -> AGENTS.md` intact. |
| 7 | `npx vitest run` | **775 passed, 1 failed** — `compose-args.test.ts`, **pre-existing**: independently reproduced on clean `upstream/development` via stash, and this PR touches no compose surface. |
| 7b | `oh init` into a temp dir | ✅ Produces no `.oh/context/`. |
| 8 | `/retro` writes no file | ✅ Report-only contract is now asserted by `retro-deterministic-contract.sh` (`ro-d2`). |
| 9 | CHANGELOG | ✅ Under `### Removed`/`### Changed`, naming each deletion with evidence and stating `REPO_MAP.md` went as *unproven*. `changelog-entry-length` PASS. |

### Red-team records

Each probe's subject was deliberately broken, then restored (`cmp`-verified) and re-confirmed green.

| Probe | Break introduced | Failure message produced |
|---|---|---|
| `context-tier-size-budget` | appended 1000 B to `AGENTS.md` | `REGRESSION: always-on context budget exceeded: AGENTS.md is 10374 B, over the 9500 B budget` |
| `context-tier-size-budget` | moved `AGENTS.md` away | `SKIPPED: the always-on context file AGENTS.md is not present` (rc=2, correct oracle state) |
| `git-skill` (A) | mangled `Default Target Branch` + `After Push` headings | `REGRESSION: git workflow skill migration incomplete: base detection moved to skill after-push CI policy moved to skill` |
| `git-skill` (B) | mangled `` Full policy: `/git` § Worktrees `` | `REGRESSION: … /worktrees points to /git` |
| `git-skill` (C) | mangled the cleanup-cron and CHANGELOG `/git` pointers | `REGRESSION: … cleanup cron points to /git CHANGELOG top pointer references skill` |
| `builder-skill-consolidation` | rewrote all 3 `` `.claude/rules/` `` occurrences | `REGRESSION: rule type omits the provider-specific rule surface` |
| `retro-deterministic-contract` | replaced the triage-tag regex with `^` | `REGRESSION: ro-f validator accepted a probe candidate with no triage tag or probe id` |

`git-skill.sh` previously asserted that `.oh/context/rules/git.md` **exists**. Rather than drop the assertion, it was retargeted to prove the git conventions live in `/git` and are advertised from `AGENTS.md`, `/worktrees`, the cleanup cron, and the CHANGELOG pointer.

### Deliberate surviving `.oh/context` mentions

- `.oh/cli/src/__tests__/init.test.ts` — **negative** assertions (`.toBe(false)`, `.not.toContain`) proving the tier is no longer shipped.
- `.oh/evals/probes/retro-deterministic-contract.sh` — an **absence guard**: it greps for the string in order to fail when present.
- `.oh/evals/capability/RESULTS.md` — the retirement annotation, which must name what was retired.
- `docs/rfcs/preserved-changelog-rationale.md` — a historical record whose own header forbids repairing quoted paths ("that would falsify the record").
- `.oh/evals/datasets/.../oracle/diff.patch` — an immutable dataset oracle snapshot.

## Defects found and fixed beyond the plan

Several probes had hard floors that the deletions would have tripped: `capability-benchmark-schema` (`>=3` CB specs, would see 2) and `datasets-schema` (`>=3` DS folders, would see 2) — both lowered to `>= 2`. `context-tier-size-budget`'s `SINGLE_FILE_SHARE_MAX=60` had to go entirely, since one file in the tier is necessarily 100% of it. Orphans left by the deletions were also swept: `repo-orientation-benchmark-score.test.ts`, `eval-quality.md`, `.oh/evals/README.md`'s `ablation` tier, and the `audit-architecture.md` wiki citation.

## Out of scope, deliberately

Root `README.md` and `docs/README.md` untouched; `.oh/README.md` received dangling-path corrections only. No nested `AGENTS.md` created. `CHANGELOG.md` history immutable.

## Known follow-ups

- Dangling `.oh/agents/advisor.md` citations across ~10 files (deliberately deferred).
- Two pre-existing dangling `.oh/memory/` references in `.oh/README.md` from #836.
- `compose-args.test.ts` fails on `development` independently of this work.


ryaneggz:
## CI green — all four checks pass

| Check | Result |
|---|---|
| Eval Probe Regression Gate | ✅ pass |
| Lint, Typecheck, Build & Test | ✅ pass |
| Validate sandbox compose and image build | ✅ pass |
| Boot Path Lint (shellcheck + hadolint) | ✅ pass |

Two guards caught real problems in this PR before CI did — worth recording, since both are cases where a green local run was not evidence.

### 1. `protected-path-deletion` did exactly its job

It fired on six deleted protected paths: the four `.oh/context/*.md` identity files, `.oh/skills/retro/scripts/check-identity-duplicates.sh`, and `.oh/skills/audit/scripts/context-audit-runner.sh`. Its design note is precise about why the sibling probe cannot see this: *"Removing the path and its list entry in the same commit keeps `protected-paths-resolve.sh` green."* That is exactly what this PR does, and `protected-paths-resolve` was indeed green throughout.

It also reads `.claude/protected-paths.txt` **at the merge base**, so amending the list in-branch cannot clear it. The only way through is the intended one: name each path and justify it. Added as [`.oh/tasks/remove-oh-context-tier/evidence.md`](https://github.com/mifunedev/openharness/blob/task/868-remove-oh-context-tier-upstream/.oh/tasks/remove-oh-context-tier/evidence.md), tracked via `git add -f`. Note the probe resolves evidence with `git show HEAD:<doc>`, so staging is not enough — it must be committed.

### 2. `audit-stale-references` caught a self-inflicted regression

The first push failed the eval gate on a file this PR itself added: the delegation artifact `.oh/tasks/remove-oh-context-tier-upstream/prompt.md` contained a bare `context-audit` token at line 816. It passed locally only because the file was untracked at the time I ran the suite — tracking it is what made the probe see it.

Since #865 deleted the accumulated `.oh/tasks/` folders wholesale, shipping a planning artifact would have cut against that decision anyway. Dropped from the commit; it remains on disk, untracked. Only `evidence.md` is tracked, because a probe requires it.

### Note on the one local test failure

`compose-args.test.ts > keeps config.json override paths as literal argv entries` fails in this sandbox but **passes in CI**, as this run shows. It reproduces on clean `development` via `git stash`, and this PR touches no compose surface — so it is a local-environment artifact, not a branch regression. The PR description called it "pre-existing"; more precisely, it is pre-existing *and* environment-specific.

PR #875, opened 2026-08-29
FROM task/874-crons-to-root TO task/872-worktrees-layout

### LINKED ISSUE #874: task(#874): move crons to repo root
Stacked on #872 / #873.

## Problem

`.oh/` is the *portable control plane* — the machinery Open Harness ships: the `oh` CLI, lifecycle scripts, installer inputs, skills, hooks. Cron definitions are not that. They are operator schedule content, authored per deployment, sitting next to `docs/` in kind rather than next to `cli/` and `scripts/`.

Same argument that moved `.worktrees/` and `projects/` to the repo root in #872.

## Proposal

`.oh/crons/` → `crons/`. `CRONS_DIR` already exists as the override, so this is a default change, not a new mechanism.

`crons/` is **tracked** content, unlike `.worktrees/` and `projects/`, so it keeps its existing `README.md` — the nested-`AGENTS.md` exemption from #872 is for git-boundary directories only, and `crons/` is not one.

## The one non-mechanical piece

`copyOhPayload` (`.oh/cli/src/lib/vendor.ts`) is hard-scoped `.oh/` → target `.oh/`, guarded by `assertDestInTarget`. A root-level `crons/` cannot ship to initialized projects through it, so `manifest.json`'s `"crons/**"` entry stops working.

Fix: add a `rootInclude` list to the manifest and a parallel `copyRootPayload` pass with its own dest guard, invoked by `oh init` / `oh update` after the `.oh` pass. The existing `.oh` guard is not loosened.

Also needs a migration warning: a stale `<target>/.oh/crons/` from a pre-move `oh init` would leave two cron dirs, and the runtime would silently scan the new empty one.

## Scope

Path resolution (`cron-runtime.ts`, `entrypoint.sh`, both compose files, `.example.env`, `restart-openharness-tmux.sh`, `.gitignore`), the `ci-harness.yml` path filters, the CLI root payload, ~8 probes and ~20 doc/skill files, and a host-side `git mv` plus a cron-runtime restart.

ryaneggz:
Closes #874

**Stacked on #873.** Review that one first; GitHub will retarget this to `development` automatically when #873 merges.

## What changed

`.oh/crons/` → `crons/`.

`.oh/` is the *portable control plane* — the `oh` CLI, lifecycle scripts, installer inputs, skills, hooks. Cron definitions are not that. They are operator schedule content, authored per deployment, closer in kind to `docs/` than to `cli/`. Same argument that moved `.worktrees/` and `projects/` out in #873.

`CRONS_DIR` already existed as the override, so the runtime side is a default change, not a new mechanism. `crons/` is **tracked** content, so it keeps its `README.md` — the nested-`AGENTS.md` exemption from #873 is for git-boundary directories, and `crons/` is not one.

## The one part that isn't mechanical

`copyOhPayload` (`.oh/cli/src/lib/vendor.ts`) is hard-scoped `.oh/` → `<target>/.oh/`, guarded by `assertDestInTarget`. A root-level `crons/` cannot ship to initialized projects through it, so `manifest.json`'s `"crons/**"` entry would have silently stopped delivering the four starter crons.

Added a parallel pass rather than loosening the existing guard:

- **`manifest.rootInclude`** — globs relative to the **repo root**, landing in `<target>/`. `exclude` applies to both lists. Optional, so an older manifest still loads.
- **`copyRootPayload`** — mirrors `copyOhPayload` with its own `assertDestInRoot` guard. It walks **only the top-level dirs its own patterns name** (via `rootPayloadDirs`), never the whole repo root — walking the root would drag in `node_modules`, `.git`, `projects/`, and every worktree.
- **`oh update` migration warning** — a stale `<target>/.oh/crons/` from a pre-move install would leave two cron dirs with the runtime silently scanning the new empty one. `oh update` now reports it and tells the operator to move local edits, rather than deleting anything.

The `.oh` guard is byte-for-byte unchanged.

## Surfaces touched

| | |
| --- | --- |
| Resolution | `cron-runtime.ts`, `entrypoint.sh`, both compose files, `.example.env` (+ template), `restart-openharness-tmux.sh` (6 `.pid`/`.cron.log` literals) |
| Ignore | `.gitignore` — `crons/.pid`, `crons/*.log` |
| CI | `ci-harness.yml` path filters — **miss this and CI stops triggering on cron edits** |
| CLI | `manifest.ts`, `vendor.ts`, `init.ts`, `update.ts`, `manifest.json` |
| Probes | 8 files, ~25 literals — including `drift-check-cron-staleness-glob.sh`, whose *glob* moved |
| Docs/skills | ~20 files, plus `.oh/README.md`'s relocation narrative and `docs/oh-directory-layout.md`'s manifest-shape row, which now document `rootInclude` |
| Guard | `.claude/protected-paths.txt` — the three protected cron files |

`git mv` throughout, so history follows the files.

## Verification

- Probe suite: **88 PASS / 0 FAIL / 5 SKIPPED**.
- `tsc --noEmit` clean; full `vitest run` → **773 passed**. The 6 `compose-args.test.ts` failures are pre-existing on unmodified `development` (verified against the clean root checkout) and pass in CI.
- CLI suite 426/426, including two new `copyRootPayload` cases: one proving only `rootInclude` dirs ship and that `assertDestInRoot` rejects an escape, one proving it is a no-op without `rootInclude`.
- **End-to-end `oh init` into a scratch dir**: `crons/{README,cleanup-tasks,eval-weekly,heartbeat,prompt-miner}.md` land at the target **root**, `<target>/.oh/crons` does not exist, and `--dry-run` previews `create crons/…` — the real check that the new payload pass actually delivers.

## Post-merge step (not done yet, and safe to defer)

Unlike #873, these are tracked files, so the merge itself performs the move. Two things remain on the host afterwards:

1. Move the untracked runtime artifacts: `mv .oh/crons/.pid .oh/crons/.cron.log crons/ && rmdir .oh/crons`.
2. Restart the cron runtime via `.oh/scripts/maintenance/restart-openharness-tmux.sh` — the running process holds the old resolved path — and confirm a fresh `BOOT` line in `crons/.cron.log`.

No interim breakage: the running sandbox has `CRONS_DIR=.oh/crons` in its environment from the pre-change compose file, so the live runtime keeps reading the old path until it is restarted.


ryaneggz:
Auto-closed when the #873 stack parent branch was deleted on merge; GitHub refuses to reopen a PR whose base ref is gone. Continued as #876 — same branch, same commits, rebased directly onto `development`.

PR #885, opened 2026-08-29
feat(sandbox): enable Claude Code todo tools by default

### LINKED ISSUE #884: Enable Claude Code todo/task tools by default in the sandbox
Claude Code 2.1.233+ stops providing `TodoWrite`/`TaskCreate`/`TaskGet`/`TaskUpdate`/`TaskList` on Opus 4.8, Sonnet 5, Fable 5, Mythos 5 and later unless a session opts in with `CLAUDE_CODE_ENABLE_TODO_TOOLS=1`.

Agent sessions in the sandbox lose the todo list, so multi-step work has no visible progress surface.

Set `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` in the sandbox compose environment (base and image-only), beside the existing `CLAUDE_DANGEROUSLY_SKIP_PERMISSIONS` and `CC_SAFETY_NET_*` entries, and guard it with a probe.

Refs: https://code.claude.com/docs/en/agent-sdk/todo-tracking

ryaneggz:
## Problem

Claude Code 2.1.233+ no longer provides `TodoWrite`, `TaskCreate`, `TaskGet`, `TaskUpdate` or `TaskList` on Opus 4.8, Sonnet 5, Fable 5, Mythos 5 and later versions of those families. A session must opt in with `CLAUDE_CODE_ENABLE_TODO_TOOLS=1`. Agent sessions in the sandbox therefore run without a todo list, and multi-step work has no visible progress surface.

Source: [Track todos](https://code.claude.com/docs/en/agent-sdk/todo-tracking) — the environment variable is the opt-in for the CLI; `todoFeatureEnabled` in `settings.json` only toggles the panel and is unrelated.

## Change

- Set `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` in the `environment:` block of `.devcontainer/docker-compose.yml` and `.devcontainer/docker-compose.image-only.yml`, beside the existing `CLAUDE_DANGEROUSLY_SKIP_PERMISSIONS` and `CC_SAFETY_NET_*` entries.
- Add an `env` block setting the same variable in `.claude/settings.json` and `.oh/templates/full/.claude/settings.json`.
- Add `.oh/evals/probes/todo-tools-default.sh` (tier A), asserting all four surfaces still carry the variable.

The two surfaces cover different sessions. Compose env reaches every process in the container regardless of cwd — worktrees under `.worktrees/`, clones under `projects/`, cron and tmux services — and takes effect on the next sandbox recreate. The settings `env` block covers a host-run session in this repo, applies without a container restart, and ships to every project scaffolded by `oh init --full`; note that per the settings docs most `env` values apply only after the folder is trusted, which is why compose is not sufficient on its own. `oh init --full` also copies `.devcontainer/` from the source checkout, so initialized projects inherit both.

## Surfaces

- **Host and sandbox:** applied in the sandbox definition; no host change.
- **Lifecycle doors:** not applicable — `make` and `oh` both reach these compose files through `.oh/scripts/docker-compose.sh`.
- **Canonical and provider surfaces:** applied — `.devcontainer/` is outside `.oh/`; the scaffold copy is the canonical `.oh/templates/full/.claude/settings.json`, and no symlink is touched.
- **Root and scaffold:** both; `oh init --full` copies the source `.devcontainer/`.
- **Interactive and headless processes:** applied to both — container environment, inherited by Herdr panes and tmux services alike.
- **Local and remote operation:** applied; the variable is in the container, not an attached shell.
- **Parallel operation:** no shared mutable state.
- **Public documentation:** not applicable — no user-facing terminology change.
- **Verification:** `todo-tools-default.sh` covers all four surfaces; full suite run below.

## Verification

```
bash .oh/skills/eval/run.sh
ran 95 probe(s) — 0 REGRESSION
todo-tools-default               PASS        new-pass  (4 surfaces checked)
```

Not verified here: the running container predates the change, so the compose variable takes effect on the next `make sandbox` recreate.

Closes #884

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,5 +1,8 @@
 {
   "enabledMcpjsonServers": ["debugmcp"],
+  "env": {
+    "CLAUDE_CODE_ENABLE_TODO_TOOLS": "1"
+  },
   "permissions": {
     "deny": [
       "Read(file_path=**/.env*)",
--- a/.devcontainer/docker-compose.image-only.yml
+++ b/.devcontainer/docker-compose.image-only.yml
@@ -50,6 +50,7 @@ services:
       # CC_SAFETY_NET_OFF left unset (kill-switch, opt-in per session).
       - CC_SAFETY_NET_STRICT=1
       - CC_SAFETY_NET_WORKTREE=1
+      - CLAUDE_CODE_ENABLE_TODO_TOOLS=1
       - GIT_USER_NAME=${GIT_USER_NAME:-}
       - GIT_USER_EMAIL=${GIT_USER_EMAIL:-}
       - GH_TOKEN=${GH_TOKEN:-}
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -75,6 +75,7 @@ services:
       # CC_SAFETY_NET_OFF left unset (kill-switch, opt-in per session).
       - CC_SAFETY_NET_STRICT=1
       - CC_SAFETY_NET_WORKTREE=1
+      - CLAUDE_CODE_ENABLE_TODO_TOOLS=1
       - GIT_USER_NAME=${GIT_USER_NAME:-}
       - GIT_USER_EMAIL=${GIT_USER_EMAIL:-}
       - GH_TOKEN=${GH_TOKEN:-}
--- a/.oh/evals/probes/todo-tools-default.sh
+++ b/.oh/evals/probes/todo-tools-default.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # tier: A
-# desc: Claude Code 2.1.233+ drops the todo/task tools on current models unless CLAUDE_CODE_ENABLE_TODO_TOOLS=1; the sandbox sets it so every agent session gets the todo list by default
+# desc: Claude Code 2.1.233+ drops the todo/task tools on current models unless CLAUDE_CODE_ENABLE_TODO_TOOLS=1; the sandbox compose env and both Claude settings files set it so every agent session gets the todo list by default
 set -euo pipefail
 
 PROBE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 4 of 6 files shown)


ryaneggz:
Closing on a four-seat council review. Superseded by #886.

## Ruling

Seats: model behavior, harness architecture, operator observability, adversarial critic. Votes were 2× CLOSE AND REPLACE, 2× SHIP NARROWED — but **all four rejected this PR as written**. The split was only over whether a narrowed one-line flag should ship beside the real fix.

Unanimous findings:

1. **The durable ledger already exists.** `.oh/tasks/<slug>/` carries `prd.json` (authoritative task graph), `progress.txt` (running log ending `STATUS: COMPLETE`) and `evidence.md` (`.oh/tasks/README.md:8-20`), gated by `/spec execute`'s evidence check (`execute.md:393-408`) and promoted into the PR body (`:412-437`). `crons/heartbeat.md:20-21` already polls `progress.txt` mtime to find stalled runs. Todo state is per-session and does not survive `/compact` (`execute.md:221`), session death, or worktree handoff. Re-enabling the tools adds a weaker in-context ledger competing with the durable one.

2. **Nothing in the harness consumes todo state.** A repo-wide grep for `TodoWrite|TaskCreate|TaskUpdate|TaskList|TaskGet` hits only this PR's own files and `.oh/skills/delegate/SKILL.md:131,172`. No hook, script, probe, statusLine or cron reads it. `notify_slack.sh:54-67` sends the last assistant text block; Herdr derives agent status from pane/terminal-title detection (`herdr/SKILL.md:80-84`). The remaining loss is the Ctrl+T panel in an attached pane — the mode `AGENTS.md` non-negotiable #3 says not to depend on.

3. **This PR's real (accidental) target is a `/delegate` bug.** `.oh/skills/delegate/SKILL.md:131` and `:172` instruct `TaskCreate`/`TaskUpdate` from a canonical skill symlinked into `.claude`, `.codex` and `.pi`. Verified in a live Opus 5 sandbox session: those tools are absent, so the step is a no-op the model cannot report.

4. **The probe was Goodharted.** Judged against this repo's own standard (`.oh/skills/audit/references/eval-quality.md`): check 6 too-narrow/special-cased (`:104-112`) — four hardcoded greps for one literal that can only fail if someone deletes the line it guards; check 4 asserts an implementation detail, not a user outcome (`:83-93`). It also carries no `# source:` header, which `.oh/evals/README.md:36-40` requires. A probe that closes no lesson is a preference in a probe's uniform.

5. **The "harness-owned ledger with Task/Todo as presentation adapters" counter-proposal is rejected** as re-proposing `.oh/tasks/`. Fails `AGENTS.md:66` (no machinery for impressiveness), `:70` (YAGNI), `:184` (smaller truthful model over a complete-looking abstraction).

## The tiebreaker

The critic's strongest point — that closing leaves `/delegate` broken today — is real, and Anthropic shipping an opt-in rather than a removal is an invitation, not a stigma. What defeats it: **the flag cannot fix `/delegate` for Codex or Pi.** Those providers never had these tools (`docs/harnesses/pi.md:39` — Pi's `Task*` is a different plugin with different semantics), so the coupling violates non-negotiable #2 with or without the env var. Only moving the wave graph into `.oh/tasks/` fixes it for every provider, and that fix needs no flag.

## Measured cost

The `TodoWrite` description block in the installed 2.1.251 bundle is ~9,055 chars ≈ 2,260 tokens — roughly the entire always-on context tier this repo caps at 9,500 B (`context-tier-size-budget.sh:13`) — before the four `Task*` schemas and per-turn reminder injection. Boundary detection was heuristic; treat as order-of-magnitude.

## Disposition

- Closed: this PR and #884.
- Opened: #886 — rewrite `/delegate` steps 4 and 5b onto `.oh/tasks/`, with a two-sided coupling probe (no `.oh/skills/**` file may instruct `Task*` unless the sandbox enables them) carrying a proper `# source:`.
- Branch `feat/todo-tools-default` stays pushed for reference; nothing merged to `development`.

Reopen signal if this ruling was wrong: a `/delegate` run on ≥3 tasks with ≥1 dependency where workers land out of order, or a `/spec execute` cron run that loses the story graph after `/compact` and cannot restore it from `progress.txt`.

PR #888, opened 2026-08-29
fix(delegate): track the wave graph in the .oh/tasks ledger

### LINKED ISSUE #886: /delegate hardcodes Claude-Code-only Task* tools in a provider-shared skill
## Defect

`.oh/skills/delegate/SKILL.md` instructs the agent to use Claude-Code-only tools:

- `:131` (step 4) — "Use `TaskCreate` for each task. Then use `TaskUpdate` with `addBlockedBy` to wire dependencies."
- `:172` (step 5b) — "update each task via `TaskUpdate`"

`.oh/skills/` is the canonical source symlinked into `.claude/skills`, `.codex/skills` and `.pi/skills`, so this is a provider-shared primitive with a single-provider dependency. That strains `AGENTS.md` non-negotiable #2 (coding-harness choice must not change the workspace).

Two independent breakages:

1. **Codex and Pi: broken already.** Neither has `TaskCreate`/`TaskUpdate`. `docs/harnesses/pi.md:39` confirms Pi's `Task*` is a different plugin with different semantics, so the names are not a shared contract.
2. **Claude Code: broken since v2.1.233 (2026-08-14).** `TodoWrite`, `TaskCreate`, `TaskGet`, `TaskUpdate` and `TaskList` are no longer provided by default on Opus 4.8, Sonnet 5, Fable 5, Mythos 5 and newer.

Verified in a live sandbox session on Opus 5: the tools are absent, so step 4 is a no-op the model cannot report. The wave graph degrades to prose held in a context window that `/spec execute` later compacts (`.oh/skills/spec/references/execute.md:221`) — during a fan-out explicitly designed not to share mutable state.

This fails silently. Waves launch without dependency wiring and workers land out of order, which reads as model flakiness rather than a missing tool.

## Fix

Rewrite steps 4 and 5b to track the wave graph in the durable ledger the harness already owns: `.oh/tasks/<slug>/` — `prd.json` for the graph, `progress.txt` for the running log (`.oh/tasks/README.md:8-20`). The topological wave plan `/delegate` already computes in-message (`SKILL.md:110-128`) is the load-bearing artifact; `TaskCreate` only mirrored it into a per-session store that does not survive compaction, session death, or handoff.

That fix is provider-neutral and needs no environment flag.

## Verification

A probe asserting the two-sided coupling invariant: no file under `.oh/skills/**` may instruct `TaskCreate`/`TaskUpdate` unless the sandbox enables the tools. It goes green legitimately once `/delegate` no longer needs them, rather than freezing a boolean.

## Context

Supersedes #884 and PR #885, which set `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` sandbox-wide. A four-seat council found that flag treats the symptom on one provider only and cannot fix Codex or Pi. See the close comment on #885.

ryaneggz:
Implements the four-seat council ruling that closed #885. Closes #886.

## Problem

`.oh/skills/delegate/SKILL.md` instructed Claude-Code-only tools from the canonical pack that is symlinked into `.claude/skills`, `.codex/skills` and `.pi/skills`:

- `:131` (step 4) — "Use `TaskCreate` for each task. Then use `TaskUpdate` with `addBlockedBy` to wire dependencies."
- `:172` (step 5b) — "update each task via `TaskUpdate`"

Two independent breakages:

1. **Codex and Pi: broken since always.** Neither provides those tools. `docs/harnesses/pi.md:39` confirms Pi's `Task*` is a different plugin with different semantics, so the names were never a shared contract. This strains `AGENTS.md` non-negotiable #2.
2. **Claude Code: broken since v2.1.233 (2026-08-14).** `TodoWrite`, `TaskCreate`, `TaskGet`, `TaskUpdate` and `TaskList` are no longer provided by default on Opus 4.8, Sonnet 5, Fable 5, Mythos 5 and newer.

Verified in a live Opus 5 sandbox session: the tools are absent, so step 4 was a no-op the model could not report. The wave graph degraded to prose held in a context window that `/spec execute` later compacts (`execute.md:221`) — during a fan-out explicitly designed not to share mutable state. It fails silently: waves launch without dependency wiring and workers land out of order, which reads as model flakiness rather than a missing tool.

## Change

**`/delegate` steps 4 and 5b now write a run ledger** under `.oh/tasks/<slug>/`:

| File | Contents |
|------|----------|
| `delegate-graph.json` | Per-task ID, title, description, `dependsOn`, files, complexity, model override + reason, thinking level, acceptance criteria, wave, `status` |
| `delegate-log.txt` | Append-only; one line per wave boundary and per status change |

- The run directory reuses the enclosing `/spec execute` slug when there is one, else `delegate-<kebab-topic>-<YYYY-MM-DD>`.
- Both files are owned by this skill. It never writes `prd.json` or `progress.txt` — those belong to the Advisor, and `.oh/tasks/README.md` says `progress.txt` must not be edited by hand.
- Step 5b persists status *before* spawning the next wave, so a crash between waves leaves a readable graph.
- **Resume rather than restart**: an existing `delegate-graph.json` is read first, and only `pending`/`FAIL`/`BLOCKED` tasks re-run.

The fix is provider-neutral and needs no environment flag — which is the point. `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` (the closed #885) would have restored the tools on Claude Code only, leaving Codex and Pi broken.

**New probe `skills-task-tool-coupling.sh`** (tier A), a two-sided invariant: a canonical skill may instruct the task tools only while the sandbox enables them, and the sandbox may enable them only while some skill needs them. It carries a `# source:` header naming the coupling, per `.oh/evals/README.md:36-40`.

This replaces the probe on the closed PR, which the council judged Goodharted against this repo's own standard (`.oh/skills/audit/references/eval-quality.md` check 6, too-narrow/special-cased; check 4, asserts an implementation detail rather than an outcome). The new probe goes green legitimately when `/delegate` stops needing the tools — which is exactly what this PR makes happen — rather than freezing a boolean forever.

## Surfaces

- **Canonical and provider surfaces:** applied — the edit is in canonical `.oh/skills/delegate/`, inherited by all three provider symlinks. No mirror patched.
- **Host and sandbox:** not applicable — no runtime or image change.
- **Lifecycle doors:** not applicable.
- **Root and scaffold:** applied to both; `.oh/skills/` ships via the manifest.
- **Interactive and headless processes:** applied to both — the ledger is on disk, so an unattended run is inspectable without an attached pane.
- **Local and remote operation:** improved; the graph now survives disconnect, session death, and worktree handoff.
- **Parallel operation:** improved; the wave graph is explicit shared state on disk rather than hidden per-session state.
- **Public documentation:** not applicable — no user-facing terminology change.
- **Verification:** below.

## Verification

```
bash .oh/skills/eval/run.sh
ran 95 probe(s) — 0 REGRESSION
skills-task-tool-coupling        PASS        new-pass
```

Both failure directions of the new probe were exercised before commit, then reverted:

- Appending a `TaskCreate` instruction to a skill with the sandbox not enabling → `REGRESSION`, exit 1, naming the file.
- Adding `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` to compose with no skill consumer → `REGRESSION`, exit 1, naming it dead configuration.

Not verified here: no live `/delegate` fan-out was run against the new ledger. The council's suggested check is a `/delegate --dry-run` on ≥3 tasks with ≥1 dependency; `--dry-run` deliberately writes no files, so a real run is needed to exercise the write path.

## Council findings not acted on

- The researcher's "harness-owned ledger with Task/Todo as presentation adapters" was **rejected** as re-proposing `.oh/tasks/`, which already carries `prd.json`, `progress.txt` and `evidence.md` behind an evidence gate (`AGENTS.md:66,70,184` — YAGNI, smaller truthful model).
- `.gitignore:20-21` gitignores `.oh/tasks/*`, so ledger durability across git depends on `git add -f`. Durable by gate, not by default. Not addressed here — worth its own issue.
- 5 of 95 probes lack the required `# source:` header. Not addressed here.

### DIFF
--- a/.oh/evals/probes/skills-task-tool-coupling.sh
+++ b/.oh/evals/probes/skills-task-tool-coupling.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+# tier: A
+# source: council review 2026-08-29 (issue #886) — /delegate instructed Claude-Code-only
+#         TaskCreate/TaskUpdate from .oh/skills/, the canonical pack symlinked into
+#         .claude, .codex and .pi. Codex and Pi never had those tools, and Claude Code
+#         2.1.233 stopped providing them by default, so the step became a silent no-op.
+# desc: the canonical skill pack and the sandbox agree about the Claude-Code-only task
+#       tools — a skill may instruct them only while the sandbox enables them, and the
+#       sandbox may enable them only while some skill needs them
+set -euo pipefail
+
+PROBE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+ROOT="$(cd "$PROBE_DIR" && git rev-parse --show-toplevel 2>/dev/null)" \
+  || ROOT="$(cd "$PROBE_DIR/../../.." && pwd)"
+
+SKILLS="$ROOT/.oh/skills"
+TOOLS='TodoWrite|TaskCreate|TaskGet|TaskUpdate|TaskList'
+
+if [[ ! -d "$SKILLS" ]]; then
+  echo "SKIPPED: .oh/skills/ absent on this branch" >&2
+  exit 2
+fi
+
+mapfile -t consumers < <(
+  grep -rlE "\`($TOOLS)\`" "$SKILLS" 2>/dev/null | sed "s|^$ROOT/||" | sort
+)
+
+enabled_in=()
+for f in "$ROOT/.devcontainer/docker-compose.yml" \
+         "$ROOT/.devcontainer/docker-compose.image-only.yml" \
+         "$ROOT/.claude/settings.json" \
+         "$ROOT/.oh/templates/full/.claude/settings.json"; do
+  [[ -f "$f" ]] || continue
+  grep -qE 'CLAUDE_CODE_ENABLE_TODO_TOOLS' "$f" && enabled_in+=("${f#"$ROOT"/}")
+done
+
+if (( ${#consumers[@]} > 0 && ${#enabled_in[@]} == 0 )); then
+  echo "REGRESSION: a canonical skill instructs Claude-Code-only task tools that the sandbox does not enable:" >&2
+  printf '  - %s\n' "${consumers[@]}" >&2
+  echo "  Those tools are absent on Codex and Pi, and on Claude Code 2.1.233+ by default." >&2
+  echo "  Fix the skill to use the .oh/tasks/ ledger, or enable CLAUDE_CODE_ENABLE_TODO_TOOLS=1 in the sandbox." >&2
+  exit 1
+fi
+
+if (( ${#consumers[@]} == 0 && ${#enabled_in[@]} > 0 )); then
+  echo "REGRESSION: the sandbox enables CLAUDE_CODE_ENABLE_TODO_TOOLS but no canonical skill needs it:" >&2
+  printf '  - %s\n' "${enabled_in[@]}" >&2
+  echo "  Dead configuration — drop it, or point at the skill that depends on the tools." >&2
+  exit 1
+fi
+
+if (( ${#consumers[@]} == 0 )); then
+  echo "PASS: no canonical skill depends on the Claude-Code-only task tools, and the sandbox does not enable them" >&2
+else
+  echo "PASS: ${#consumers[@]} skill(s) instruct the task tools and the sandbox enables them (${enabled_in[*]})" >&2
+fi
+exit 0
--- a/.oh/skills/delegate/SKILL.md
+++ b/.oh/skills/delegate/SKILL.md
@@ -45,7 +45,7 @@ flowchart TD
 
     B -->|Yes| C["Step 2: Deep-think task decomposition"]
     C --> D["Step 3: Build dependency graph"]
-    D --> E["Step 4: Create tasks + compute waves"]
+    D --> E["Step 4: Write run ledger to .oh/tasks/"]
     E --> F{--dry-run?}
     F -->|Yes| DRY["Report: task graph + wave plan"]
     DRY --> MEM_DRY[Memory Protocol]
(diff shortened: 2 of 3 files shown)


ryaneggz:
## Ledger exercised end to end

The verification gap noted in the PR body is now closed. A real `/delegate` fan-out ran against the new ledger — 4 tasks, 2 waves, one dependency edge, max parallelism 3, 0 failed, 0 blocked.

The ledger behaved as designed: `delegate-graph.json` and `delegate-log.txt` were written before any worker spawned, updated at every status transition, and persisted the wave boundary. T1 finished before T3, but T4 correctly held until wave 1 closed. The whole run is reconstructable from disk with no conversation context.

Three findings from the run:

**1. Skills resolve from the main checkout, not the active worktree.** Invoking `/delegate` from a worktree loaded `/home/sandbox/harness/.claude/skills/delegate` — the pre-fix copy still instructing `TaskCreate` at step 4. The rewritten step 4 was followed deliberately, from the branch file. This is a general limitation on testing skill edits: a skill change on a branch is not exercised by invoking that skill, in any worktree, until it lands. Worth knowing for anyone reviewing skill PRs.

**2. The gitignore gap is real but should probably stay open.** `git check-ignore -v` confirms both ledger files match `.gitignore:20`, and no existing force-add step names them — `execute.md:124` stages the task directory once at scaffold, before `/delegate` writes anything, and a standalone run has no `/spec execute` around it at all. Durability by case: (a) session death, same machine — DURABLE, the resume path works; (b) fresh clone or CI — absent; (c) PR review — absent.

The audit argued against whitelisting them, and I find it convincing: this is run state, its value is exactly the (a) case, and whitelisting would dirty the tree on every delegation for a diff that decides nothing. The reviewer’s need in (c) is already served by `evidence.md`, which is on the force-add list and gated at `execute.md:400-405`. If a delegation’s structure matters to a reviewer, summarize the graph into `evidence.md` rather than shipping the raw ledger. `SKILL.md:152-153` currently leaves this to judgement without saying so — pointing that line at `evidence.md` is the smaller, better change. Not made here; flagging for reviewer preference.

**3. This PR edits a file with a known dangling reference.** `delegate/SKILL.md:290-293` lists four `.claude/agents/*.md` paths whose definitions were deleted in #865, while `:295` directly below already gives the correct instruction to use built-in `subagent_type` values. The table contradicts the line under it. Filed as #891 rather than folded in here, to keep this PR to the ledger change.

Also filed from the same run: #890 (`/interview` depends on `AskUserQuestion` with no fallback — structurally the same defect as #886) and #889 → PR #892 (four probes missing the required `# source:` header).

ryaneggz:
Rebased onto current `development` (six merges since this branch opened). Now MERGEABLE.

One conflict, in `CHANGELOG.md` only: both entries had merged into the released `## [0.5.0]` section. Moved to `## [Unreleased]`, the probe under `### Added` and the `/delegate` change under `### Fixed`. No change to `SKILL.md` or the probe itself.

Verified: `skills-task-tool-coupling.sh` PASSes against the rebased tree, and `skill-paths.sh` and `context-tier-size-budget.sh` (9480 B of a 9500 B budget) both still PASS.

Issue #891, opened 2026-08-29
Eleven skill references point at .claude/agents/ definitions deleted in #865


ryaneggz:
Two canonical skills reference agent definition files that do not exist.

`.claude/agents` is a tracked symlink to `../.oh/agents`, and that directory now contains only `README.md`, which states no agent is defined. The definitions were deleted in #865 and the directory restored as an empty pack in #866; neither PR updated the skills that cite them.

**Dangling references (11):**

- `.oh/skills/strategic-proposal/SKILL.md:126-130,138,150,162,209-215` — `expert-product`, `expert-docs`, `expert-security`, `expert-registry`, `expert-agent-systems`, `strategic-council`, `strategic-critic`. The skill's entire five-expert fan-out points at nothing.
- `.oh/skills/delegate/SKILL.md:290-293` — `implementer`, `critic`, `pm`, `council`.

Two distinct severities. `strategic-proposal` cannot execute its main flow as written. `delegate` is merely contradictory: its "Key Resources" table lists the four dangling paths, while `:295` immediately below already gives the correct instruction (use the built-in `general-purpose` type). The table is dead weight actively disagreeing with the line under it.

## Fix

Rewrite both tables to name built-in `subagent_type` values instead of file paths, following `delegate/SKILL.md:295`. Restoring definitions under `.oh/agents/` is the alternative, but #866 made the empty pack deliberate, so pointing at built-ins matches the decision already taken.

## Related, found in the same sweep

- `.oh/skills/eval/SKILL.md:25,26,27,35` and `.oh/skills/benchmark/SKILL.md:70` invoke `bash .claude/skills/eval/run.sh` — running the canonical runner through a provider mirror, which `AGENTS.md` warns against. The rest of the pack already uses `.oh/skills/eval/run.sh` (`sync/SKILL.md:86,105`, `spec/references/execute.md:266`, `audit/references/implementation.md:104`); these five lines are the outliers.
- `.oh/skills/prompt-miner/references/pi-parity.md:11-20` asserts `.pi/skills` resolves to `../.claude/skills`. It actually resolves to `../.oh/skills`, so the document's reasoning chain is false.
- `.oh/skills/imagine/SKILL.md:5` cites `.gitignore:51` for a rule now at `:47`. Behavior is unaffected; the citation is stale.

ryaneggz:
Fixed by #931 (`cdc34efb`, `feat: make skills the role primitive and add /architect`).

[...]

Issue #890, opened 2026-08-29
/interview depends on Claude-Code-only AskUserQuestion with no fallback


ryaneggz:
Same defect class as #886, found by a coupling sweep of the canonical pack.

`.oh/skills/interview/SKILL.md` is built on `AskUserQuestion` and `ExitPlanMode` (`:15,22,32,53,85,107,147,179,189,193,195`). Both are Claude Code built-ins. `.oh/skills/` is the canonical pack symlinked into `.claude/skills`, `.codex/skills` and `.pi/skills`, so `AGENTS.md` non-negotiable #2 requires the skill to work on all three.

On Codex and Pi, step 4 — "issue one batched `AskUserQuestion` call" — cannot execute, and the skill offers no prose-question fallback. Its only step is its coupling.

This is more load-bearing than the `/delegate` case that prompted #886: there, `TaskCreate` mirrored a graph the skill computed anyway, so losing it degraded tracking. Here the tool *is* the skill.

## Fix

Give the skill a harness-neutral path: ask the questions in prose, numbered, when the structured tool is unavailable. Phrase the instruction as "ask the user (use `AskUserQuestion` where available)" rather than naming the tool as the only route — the same shape `health-check/SKILL.md:227` needs for its single approval gate.

## Verification

`skills-task-tool-coupling.sh` (added in #888) covers only `TodoWrite`/`Task*`. Widening it to the full set of Claude-only built-ins would catch this class, but note the probe asserts a two-sided sandbox/skill agreement that does not map cleanly onto tools with no enabling flag — a separate assertion is probably cleaner than overloading that one.

PR #892, opened 2026-08-29
fix(evals): record the source lesson on five headerless probes

### LINKED ISSUE #889: Four probes ship without the required # source: header
`.oh/evals/README.md:36-40` requires every probe to carry a `# source:` line naming the lesson or rule it closes. Four of 95 do not:

- `.oh/evals/probes/compose-config-path-parity.sh`
- `.oh/evals/probes/env-schema-parity.sh`
- `.oh/evals/probes/harness-yaml-migration.sh`
- `.oh/evals/probes/oh-init-headless-config.sh`

Found with `grep -L "^# source:" .oh/evals/probes/*.sh`. All four also show `?` in the `source` column of `.oh/evals/RESULTS.md`, so the runner already surfaces the gap.

This is an origin gap, not a regression: each shipped with only `tier` and `desc`. The `b4ea86f1` comment-strip preserved every `tier`/`source`/`desc` header it found.

All four trace to merged PRs read in full — #833 (three of them) and #827 (retargeted by #833). A probe with no recorded source is a probe nobody can retire, because no one can tell what it was protecting.

ryaneggz:
Closes #889.

## Problem

`.oh/evals/README.md:36-40` requires every probe to carry a `# source:` line naming the lesson or rule it closes. Four of 95 did not, and surfaced as `?` in the `source` column of `.oh/evals/RESULTS.md` — the runner was already reporting the gap.

A probe with no recorded source is a probe nobody can retire: there is no way to tell what it was protecting, so it accumulates forever. `.oh/evals/README.md:80` is explicit that not every lesson becomes a probe, which only works if the ones that exist say why.

This is an origin gap, not a regression. Each of the four shipped with only `tier` and `desc`; the `b4ea86f1` comment-strip preserved every `tier`/`source`/`desc` header it encountered.

## Change

One header line per probe, each citing a merged PR read in full:

| Probe | Source |
|---|---|
| `compose-config-path-parity.sh` | PR #833 — the wrapper and VS Code "Reopen in Container" paths must resolve the same service |
| `env-schema-parity.sh` | PR #833 — `.devcontainer/.example.env` is the one schema; five vars were consumed but undocumented |
| `harness-yaml-migration.sh` | PR #833 — `migrate-harness-yaml.sh` append / uncomment-in-place / preserve / overwrite, plus a silent no-op second run |
| `oh-init-headless-config.sh` | PR #827 — installer answers landed in the losing config file; retargeted to the `.example.env` template by #833 |

These are PR numbers rather than issue numbers because both are squash-merge subjects and neither commit body carries a `Closes`/`Fixes` line, so no underlying issue is discoverable from the repo. The existing corpus already cites non-issue artifacts where that is what exists.

Header comments only. No logic, assertion, `set -euo pipefail`, `# desc:` line or exit code was touched.

## Verification

Per-probe exit codes captured before and after the edit:

| Probe | Before | After |
|---|---|---|
| compose-config-path-parity | 0 | 0 |
| env-schema-parity | 0 | 0 |
| harness-yaml-migration | 0 | 0 |
| oh-init-headless-config | 2 (SKIPPED) | 2 (SKIPPED) |

- `grep -L '^# source:' .oh/evals/probes/*.sh` → empty.
- `bash .oh/skills/eval/run.sh` → 94 probes, runner exit 0, **0 REGRESSION**.
- `RESULTS.md` shows the four `source` columns populated, confirming the header is machine-read rather than decorative.

## Provenance

Found and fixed by a `/delegate` fan-out run to exercise the run ledger added in #888 — four tasks, two waves, dependency-ordered. The inventory task corrected the count from an earlier estimate of five to the actual four. Two further findings from the same run are filed separately as #890 and #891; neither is touched here.

### DIFF
--- a/.oh/evals/probes/compose-config-path-parity.sh
+++ b/.oh/evals/probes/compose-config-path-parity.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # tier: A
+# source: PR #833 (remove harness.yaml — the wrapper and VS Code "Reopen in Container" paths must resolve the same service) 2026-08-26
 # desc: the wrapper path and the VS Code "Reopen in Container" path resolve the same service — the parity harness.yaml made impossible
 set -euo pipefail
 
--- a/.oh/evals/probes/config-schema-parity.sh
+++ b/.oh/evals/probes/config-schema-parity.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # tier: A
+# source: PR #833 (one schema file — DOCKER_SOCKET, SANDBOX_SSH, OH_SANDBOX_IMAGE, OH_PULL_POLICY, SKIP_PNPM_INSTALL were consumed but undocumented); rewritten for the oh.json/secrets split by PR #887
 # desc: the oh.json/.env split loses no variable — every compose-interpolated var is either a documented oh.json field or an allow-listed secret, and neither surface holds the other's keys
 set -euo pipefail
 
--- a/.oh/evals/probes/harness-yaml-migration.sh
+++ b/.oh/evals/probes/harness-yaml-migration.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # tier: A
+# source: PR #833 (migrate-harness-yaml.sh — append / uncomment-in-place / preserve / overwrite, plus a silent no-op second run) 2026-08-26
 # desc: migrate-harness-yaml.sh carries a live harness.yaml into the oh.json/root-dotenv split — non-secret settings land as oh.json fields (never .devcontainer/.env or the retired .oh/config.json), the file is renamed, and the second run is a no-op
 set -euo pipefail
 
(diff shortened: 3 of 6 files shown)


ryaneggz:
Rebased onto current `development`, which had taken six merges (#882, #883, #887, #893, #894, #895) since this branch opened. Now MERGEABLE.

Three conflicts, all caused by the one-door migration moving the target set:

- **`env-schema-parity.sh` was deleted** and replaced by `config-schema-parity.sh` (#887). Its `# source:` header moved to the successor, with the text updated to say the schema was rewritten for the `oh.json`/secrets split rather than describing the retired single-file schema.
- **`harness-yaml-migration.sh`** and **`oh-init-headless-config.sh`** were both retargeted, so their `# desc:` lines had changed underneath the branch. Kept `development`'s current descriptions and added the `# source:` line above each; the source line on `oh-init-headless-config.sh` now records both retargets (#833, then #887).
- **`CHANGELOG.md`** merged the entry into a released section. Moved to `## [Unreleased]` under `### Fixed`.

**Scope grew by one.** #887 added `oh-config-surfaces.sh` after this branch opened, and it also shipped without a `# source:` header — so the same defect the branch closes had already reopened. It is fixed here too; the title and CHANGELOG entry now say five probes, not four.

Verified: `grep -c "^# source:"` is non-zero for every file in `.oh/evals/probes/`. Four of the five run PASS locally; `oh-init-headless-config.sh` exits 2 (SKIPPED) because `.oh/cli/dist/oh.js` is not built in this worktree, which is its documented skip path, not a regression.

PR #897, opened 2026-08-29
feat(cli): support T3 Code mobile access over Tailscale

### LINKED ISSUE #858: feat(cli): support T3 Code mobile access over Tailscale
## Summary

Add Tailscale as an optional OpenHarness CLI-managed tool and provide a complete, secure path for connecting the T3 Code mobile app to T3 Code running headlessly in a remote OpenHarness environment.

This issue also owns the corresponding documentation update in `mifunedev/openharness-web`.

## Problem

The current T3 Code docs tell users to launch `npx t3` and suggest Cloudflared when sharing beyond the attached host. That does not clearly explain the supported mobile/remote flow.

Upstream T3 Code now distinguishes:

- `npx t3` — normal/local launch
- `npx t3 serve` — explicit headless server
- `npx t3 pair` — mint a fresh pairing token for an already-running server
- `--tailscale-serve` / `pair --tailscale` — advertise the backend through Tailscale Serve HTTPS
- desktop-managed SSH launch — desktop-only convenience, not the mobile connection path

For the T3 Code mobile app, the remote environment must expose a reachable authenticated backend. Tailscale is the preferred default because it avoids publishing port 3773 to the public internet, but today OpenHarness neither installs/configures it through `oh` nor explains which machine/network namespace must run it.

The present T3 page also says restarting T3 Code is required to mint another token; upstream now supports `npx t3 pair`.

## Desired operator experience

A user with OpenHarness on a remote machine should be able to:

```bash
oh tool install tailscale
# authenticate/join the remote environment to the user's tailnet
/t3 start --tailscale
```

Then:

1. Install/sign in to Tailscale on the phone.
2. Join the phone to the same tailnet.
3. Open the T3 Code mobile app.
4. Scan or paste the pairing URL printed by T3 Code.
5. Reconnect later using the saved authenticated session without exposing T3 Code publicly.

For an already-running server, OpenHarness should surface the equivalent of:

```bash
npx t3 pair --tailscale
```

### 1. Add Tailscale to `oh tool`

Follow the existing `agent-browser` persist-first pattern where it applies:

- Add a `tailscale` entry to the tool catalog.
- Support `oh tool install tailscale`.
- Persist the opt-in in `harness.yaml` (for example, `install.tailscale: true`) so it survives container recreation.
- Live-install into a running sandbox when possible.
- Make installation idempotent.
- Add presence and verified version probes for `oh tool list/status`.
- Add installer/config mappings and tests equivalent to the existing optional-tool coverage.
- Document any rebuild/restart required to activate networking.

Tailscale is not only a binary: it requires a daemon/network boundary and authentication. Installation is not complete if `tailscale version` works but T3 Code cannot create and advertise a reachable Serve endpoint.

### 2. Define the supported network boundary

Choose and implement one canonical architecture for the initial release:

- Tailscale runs inside the sandbox in a supported mode; or
- Tailscale runs on the remote host and OpenHarness safely routes the sandbox T3 port through the host tailnet.

Document the choice explicitly. Include the required Docker capabilities, `/dev/net/tun`, userspace-networking mode, socket mounts, and/or loopback port mapping as applicable.

Do not silently grant broad container privileges. Any capability/socket/port change must be opt-in and visible in generated configuration.

### 3. Add a headless T3 mobile-access flow

Update the `/t3` skill and related scripts/docs to distinguish local and remote launches.

Expected behavior:

- Headless remote launch uses `npx t3 serve`, not ambiguous `npx t3`.
- Tailscale mode uses `npx t3 serve --tailscale-serve` or an architecture-equivalent safe wrapper.
- The long-running process remains managed under the existing tmux convention.
- `/t3 url` or an equivalent command prints the current reachable pairing URL.
- A command for an already-running server uses `npx t3 pair --tailscale`.
- Missing Tailscale, unauthenticated tailnet state, incompatible Node, failed Serve configuration, and unreachable endpoints produce actionable errors.
- Verify the installed Node version satisfies T3 Code's current requirement (`^22.16 || ^23.11 || >=24.10`).
- Clarify that provider authentication (Codex/Claude/OpenCode) is separate from T3 pairing.

### 4. Update OpenHarness documentation

Update at least:

- `.oh/docs/harnesses/t3code.md`
- `.oh/docs/connecting.md`
- CLI/tool installation reference and `harness.yaml.example`

The docs must include:

- when to use `npx t3`, `npx t3 serve`, and `npx t3 pair`
- remote host and phone prerequisites
- Tailscale installation and login/join steps
- requirement that the phone and remote environment share a tailnet
- QR/pairing-link flow in the T3 Code mobile app
- reconnect and new-device pairing behavior
- background/tmux lifecycle
- troubleshooting for Tailscale state, Serve HTTPS, Node version, and port reachability
- architecture-specific explanation of where Tailscale runs

### 5. Reflect the guidance in `mifunedev/openharness-web`

Update the rendered documentation so the website matches the canonical OpenHarness guidance. Link the web PR back to this issue and avoid leaving the web copy with the older `npx t3`/Cloudflared-only instructions.

## Security requirements

- Do not bind T3 Code to a public interface by default.
- Do not enable Tailscale Funnel by default; private Tailnet access is the default.
- Treat pairing URLs/tokens as secrets and avoid persisting them in tracked files or normal logs beyond what T3 itself requires.
- Never print or commit a reusable Tailscale auth key.
- Prefer interactive `tailscale up` for a human setup; if auth-key automation is supported, accept it only through an existing secret channel.
- Document how to revoke both T3 sessions/pairing credentials and Tailscale device access.

## Acceptance criteria

- [ ] `oh tool list` and `oh tool status tailscale` report accurate enabled, installed, and version state.
- [ ] `oh tool install tailscale` persists the configuration and installs/configures the supported runtime path without requiring a manual package recipe.
- [ ] A clean remote OpenHarness environment can join a tailnet and expose T3 Code privately without a public port.
- [ ] `/t3 start --tailscale` (or the final documented equivalent) launches headless T3 Code and produces a usable mobile pairing URL.
- [ ] `npx t3 pair --tailscale` or an OpenHarness wrapper can pair an additional device without restarting the running server.
- [ ] A phone on the same tailnet can pair in the T3 Code mobile app and reconnect successfully.
- [ ] A phone outside the tailnet cannot reach the backend unless a separate explicit exposure mechanism was enabled.
- [ ] The T3 process survives shell/SSH disconnects through the existing tmux lifecycle.
- [ ] Missing/unhealthy Tailscale and incompatible Node versions fail with actionable guidance.
- [ ] Unit/catalog/config drift tests cover the new optional tool and T3 launch behavior.
- [ ] Core docs are corrected and include an end-to-end mobile recipe.
- [ ] A linked `mifunedev/openharness-web` PR publishes matching guidance.

## References

- T3 Code remote access: https://github.com/pingdotgg/t3code/blob/main/docs/user/remote-access.md
- Existing OpenHarness T3 docs: `.oh/docs/harnesses/t3code.md`
- Existing tool pattern: `.oh/cli/src/commands/tool.ts`
- Existing optional-tool key: `install.agent_browser` in `harness.yaml.example`

ryaneggz:
Closes #858.

## Architecture decision

`tailscaled` runs **inside the sandbox in userspace-networking mode** as the unprivileged `sandbox` user (`--tun=userspace-networking`, socket and state under `$HOME/.tailscale`). T3 Code stays bound to container loopback; `t3 serve --tailscale-serve` advertises it over tailnet HTTPS.

This grants **zero** new privileges: no `NET_ADMIN`, no `/dev/net/tun`, no `privileged`, no published port, no overlay file. The compose surface gains exactly one env passthrough (`INSTALL_TAILSCALE`) and one named state volume.

Rejected alternative: `tailscaled` on the remote host with the sandbox port routed through the host tailnet. It forces publishing 3773, breaks `oh tool`'s in-sandbox contract, and splits the network boundary across two machines.

## What changed

**Optional tool.** `install.tailscale` in `oh.json` maps to `INSTALL_TAILSCALE`, wired through `oh-config`, `env-file`, `config-render`, the `oh init` wizard, and `oh tool install/list/status`. The catalog entry mirrors `agent-browser`'s persist-first shape; `commands/tool.ts` needed no change. The entrypoint installs a pinned Tailscale 1.102.3 tarball with per-arch sha256 verification, is idempotent, and **never** starts the daemon or reads an auth key.

**T3 skill.** `/t3 [start|status|url|pair|logs|stop|attach|doctor|help]` with `--tailscale`, `--tailscale-port`, `--session`, `--port`, `--log`.

- `start` -> `npx --yes t3 serve`; `start --tailscale` -> `npx --yes t3 serve --tailscale-serve`
- `pair` / `pair --tailscale` adds a device against a **running** server, no restart
- `url` prints the current reachable pairing URL, `.ts.net` included
- `doctor` runs first on every `start` and fails with actionable guidance on: Node outside `^22.16 || ^23.11 || >=24.10`, missing `tailscale` binary, `tailscaled` unreachable, `NeedsLogin`, and an unreachable T3 port
- The server stays under the existing tmux convention, so it survives disconnect

Bare `npx t3` is documented as the local desktop launcher and is no longer used for headless launches.

**Docs.** `docs/harnesses/t3code.md` rewritten; `docs/connecting.md` gains an end-to-end mobile recipe (prerequisites, `tailscale up`, phone join, QR pairing, reconnect, lifecycle, troubleshooting, revocation, and where Tailscale runs and why); plus rows in `docs/installation.md`, `docs/configuration.md`, and a Caveat 4 in `docs/security-considerations.md`. Two stale claims corrected: T3 does not bind `0.0.0.0:3773`, and minting a second token no longer requires a restart. Cloudflared stays documented as the *public* preview tool, not the mobile path.

## Security

Funnel is never enabled and no `funnel` token exists under `.oh/skills/t3/` or `.devcontainer/`. Nothing binds `0.0.0.0`. Interactive `tailscale up` is the documented setup; no reusable auth key is committed or logged. Both revocation paths — T3 pairing credentials and Tailscale device access — are documented.

## Verification

- `pnpm run typecheck` clean; `pnpm run build:harness` clean
- `npx vitest run` — 947 passed. 8 failures in `.oh/scripts/__tests__/compose-args.test.ts` are **pre-existing and environment-dependent** (an ambient `docker-compose.ssh.yml` overlay); re-proved byte-identical against a clean `git archive HEAD` tree, both in failure set and failure mode, before and after the `docker-compose.yml` edit
- `bash .oh/skills/eval/run.sh` — 103 probes, **99 PASS / 4 SKIPPED / 0 REGRESSION**. The 4 skips are pre-existing and environmental
- `shellcheck -S warning` clean on every touched shell file

Two new probes, each break-tested (property broken -> REGRESSION observed -> restored):

- `tailscale-tool-boundary.sh` — 12 breaks caught: published 3773, `cap_add`, `devices`, a daemon start, a `tailscale up`, a per-arch pin disagreement, a catalog/entrypoint sha mismatch, a removed `sha256sum -c`, a Dockerfile build arg, a flipped catalog `kind`, a Funnel invocation, a committed `tskey-auth-` literal
- `t3-headless-launch.sh` — 7 breaks caught, including bare `t3`, a broken `--tailscale-serve` mapping, a dropped tmux `-d`, and a `--host 0.0.0.0`

## Not verified here — manual acceptance still required

This environment has no tailnet, no auth key, and no phone. The following acceptance criteria rest on the upstream T3 docs and are **documented-only**:

- a real tailnet join, and that `t3 serve --tailscale-serve` actually publishes on 443 and prints a `.ts.net` URL
- that `t3 pair` mints a token against a live server, and that the `url` regex matches T3's real output
- phone pairing and reconnect in the T3 Code mobile app
- that `BackendState` parsing matches real `tailscale status --json`
- the arm64 sha256 came from upstream's published `.sha256`, not a byte-for-byte local download (amd64 was downloaded and verified)
- **no `tailscale` binary was ever executed in this environment.** Every `tailscale` / `tailscaled` invocation in the docs, the skill, and the probes is stubbed or read, never run. CLI-to-daemon reachability of the documented invocation is therefore unexercised — that is exactly where the original `--socket` blocker hid, and it is closed by inspection of upstream's default socket path, not by a live round trip.

"A phone outside the tailnet cannot reach the backend" is discharged **structurally** by `tailscale-tool-boundary.sh` — no published port, no capabilities — rather than by a live negative test.

## Follow-up

Scope 5 of the issue (matching guidance in `mifunedev/openharness-web`) is **not** in this PR. That repo is a separate checkout on its own branch and needs its own PR linking #858.


ryaneggz:
## Audit — issue #858 / PR #897

Adversarial audit against the task PLAN and the issue's acceptance criteria. Every claim below was read out of the files, not taken from the executor reports or this PR body. Probes were mutation-tested, and both Tailscale sha256 pins were checked against upstream.

**Verdict: PASS-WITH-FOLLOWUPS — one BLOCKER must land before merge.** The plumbing, tests, docs, and probes are genuinely good work. One cross-workstream seam breaks the primary end-to-end flow.

### What verified clean

- **Pins are real.** `curl https://pkgs.tailscale.com/stable/tailscale_1.102.3_{amd64,arm64}.tgz.sha256` returns exactly the two literals in `.devcontainer/entrypoint.sh` and `.oh/cli/src/lib/tools/catalog.ts`. Version and both shas agree across entrypoint / catalog / test / probe. No drift.
- **`install.tailscale` → `INSTALL_TAILSCALE`** is consistent across `oh-config.ts`, `env-file.ts`, `config-render.ts`, `init.ts`, `oh.json`, `install.sh`, `migrate-harness-yaml.sh`, `docker-compose.yml`, and `docs/configuration.md`. No key-name drift.
- **Boundary held.** Compose gains exactly one env passthrough and one named volume. No `cap_add`, `devices:`, `privileged:`, no `ports:` for 3773, no `devcontainer.json` change, no overlay. The entrypoint never runs `tailscaled` and never reads an auth key.
- **Probes are load-bearing, not theatre.** Mutation-tested in a scratch tree: publishing 3773 → REGRESSION; one catalog sha flipped by a character → REGRESSION naming both sides; neutering the doctor's tailscale check → REGRESSION. `t3-headless-launch.sh` actually *executes* the script under stubbed `tmux`/`node`/`npx`/`tailscale` and asserts the captured launch string — that is a real oracle. The `tskey-auth-` scan is written so the probe's own regex does not self-match.
- **AGENTS.md:** no explanatory comments added to tracked code; canonical `.oh/` sources edited (the `.claude/skills/t3` mirror resolves to them); both long-running processes are tmux-detached.
- **WS4's judgment call is right.** Dropping bare `npx t3` from the script loses nothing: `t3 serve` is the same server, `print_summary` still advertises `http://localhost:3773`, and the container has no browser for the GUI launcher to open. Keeping bare `t3` as documented desktop guidance is the correct split.

---

### BLOCKER 1 — the custom `tailscaled` socket makes the documented flow non-functional

`.oh/skills/t3/scripts/t3-code.sh:108,137,143` · `docs/connecting.md:165,172,180,236,237,247-253` · `.oh/skills/t3/references/tailscale-mobile.md:32`

The daemon is documented and hinted as `tailscaled --socket=$HOME/.tailscale/tailscaled.sock ...`, but every *consumer* then calls the CLI bare.

Upstream ground truth: `paths.DefaultTailscaledSocket()` returns `/var/run/tailscale/tailscaled.sock` on Linux, and `cmd/tailscale/cli/cli.go` takes that default with **only** a `--socket` flag to override it — there is no `TS_SOCKET` environment variable. So a daemon on a custom socket path is invisible to:

1. `t3-code.sh:137` `tailscale status --json` → the preflight always reports *"tailscaled is not running or its socket is unreachable"*, so **`/t3 start --tailscale` can never pass `doctor`** even on a correctly configured node;
2. every bare `tailscale status` / `tailscale serve --https=443 off` / `tailscale logout` / `tailscale serve status` in the docs and revocation section;
3. T3 Code's own `--tailscale-serve` implementation, which drives Tailscale through the same default socket.

The files also disagree with each other: `docs/connecting.md:172` says `tailscale --socket=... up` while `references/tailscale-mobile.md:35` says bare `tailscale up`. Both cannot be right.

**Minimal fix — stop using a custom socket:**
- `.devcontainer/entrypoint.sh:552`, next to the existing `install -d`: add `install -d -o sandbox -g sandbox -m 0755 /var/run/tailscale 2>/dev/null || true` (recreated on every boot, so the tmpfs lifetime is fine).
- Mirror the same line into `installArgv` in `.oh/cli/src/lib/tools/catalog.ts` — `tool-catalog.test.ts` asserts entrypoint/catalog parity, so the two must move together.
- Document the daemon as `tailscaled --tun=userspace-networking --statedir=$HOME/.tailscale` and drop `--socket` from `docs/connecting.md:165,172`, `references/tailscale-mobile.md:32`, and the `tailscaled_hint()` at `t3-code.sh:108`. Every bare `tailscale ...` call then works, and so does T3's own Serve call.

Bolting `--socket` onto the doctor instead would fix (1) and (2) but not (3), so it is not a real fix.

### SHOULD-FIX 1 — the Node-floor assertion is Goodhartable

`.oh/evals/probes/t3-headless-launch.sh:44` proves the floor with `grep -qF '22.16'`, which the help text alone satisfies. Proof: replacing `elif ! node_version_ok; then` with `elif false; then` in `t3-code.sh:130` leaves the probe **PASSing**. Every other assertion in this probe is behavioral; this one is not.

**Fix:** add a third stub dir whose `node` prints `v22.15.0`, run `doctor`, and assert a non-zero exit whose output names the range.

### SHOULD-FIX 2 — the pairing token lands in a world-readable file

`.oh/skills/t3/scripts/t3-code.sh:329` — `: > "$LOG"` creates `/tmp/agent-t3code.log` at the default umask (0644) and then tees T3's stdout into it, which is exactly where the one-time pairing URL appears. The PR correctly says "treat pairing URLs as secrets"; the log does not.

**Fix:** `(umask 077; : > "$LOG")`.

### NITs

1. `.oh/evals/probes/tailscale-tool-boundary.sh` guards `cap_add` / `devices:` / `privileged:` / `ports:` but not `network_mode: host`, `sysctls:`, or `devcontainer.json`'s `appPort` / `forwardPorts` / `runArgs` — alternate ways to reintroduce exactly the exposure this probe exists to prevent.
2. `t3-code.sh:107` `tailscaled_hint()` uses `cat <<HINT`; under a minimal `PATH` it dies with `cat: command not found` (observed live while mutation-testing). Use `printf`.
3. `docs/connecting.md:163` starts the daemon without teeing a log, while `references/tailscale-mobile.md:31` tees to `/tmp/agent-tailscaled.log` and the session table names that file. Align them.
4. `.oh/scripts/install.sh:444` — column alignment of the new `_opt_install TAILSCALE` line does not match its neighbours.

### On the PR body's "Not verified here"

Honest and unusually thorough — but incomplete in the one place that mattered. It does not disclose that **no `tailscale` binary was ever executed in this environment**, so the reachability of the *documented* daemon invocation from the CLI was never exercised. That is the exact gap BLOCKER 1 sits in. Suggest adding a line to that effect alongside the fix.

### Acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | `oh tool list` / `status tailscale` report enabled, installed, version | **MET** — catalog entry + `tool.test.ts` JSON status case; `commands/tool.ts` needed no change |
| 2 | `install` persists config and installs the runtime path | **MET** — `--persist-only` and live-install cases; pinned, checksummed, idempotent, root-installed |
| 3 | Clean environment joins a tailnet and exposes T3 privately | **NOT MET as documented** — BLOCKER 1; the recipe cannot complete as written |
| 4 | `/t3 start --tailscale` produces a usable pairing URL | **NOT MET as shipped** — the preflight blocks it (BLOCKER 1); the launch string itself is proven correct by probe |
| 5 | `t3 pair --tailscale` adds a device without restart | **MET-BY-DOCS-ONLY** — `pair` action implemented and probe-asserted; upstream behavior unexecuted |
| 6 | Phone pairs and reconnects | **MET-BY-DOCS-ONLY** — no phone here; disclosed |
| 7 | Phone outside the tailnet cannot reach the backend | **MET** structurally — no published port, no capability, loopback-only; mutation-proven probe |
| 8 | Survives shell/SSH disconnect | **MET** — `tmux new-session -d` asserted from a captured launch string, not a grep |
| 9 | Missing/unhealthy Tailscale and bad Node fail with actionable guidance | **MET** for the missing-binary and unreachable-daemon paths (executed by probe); Node path implemented but only string-asserted (SHOULD-FIX 1) — and the unreachable-daemon error currently fires as a false positive (BLOCKER 1) |
| 10 | Unit/catalog/config drift tests cover the tool and launch behavior | **MET** — 6 test files + 2 probes; catalog/entrypoint parity enforced on version *and* both shas |
| 11 | Docs corrected with an end-to-end mobile recipe | **MET** apart from the socket lines in BLOCKER 1; the two stale claims (`0.0.0.0:3773`, restart-to-mint) are genuinely corrected |
| 12 | Linked `openharness-web` PR | **NOT MET, correctly deferred** — separate repo, disclosed as follow-up |

Fix BLOCKER 1 and criteria 3, 4, and 9 close.

ryaneggz:
## Re-audit of `5a3be14b` — all three findings closed

Delta-only audit. Every claim re-verified against the files and against upstream Tailscale source; probes re-run and re-broken.

**Verdict: PASS. Merge-ready.**

### BLOCKER 1 (socket) — FIXED, and the mechanism checks out

- `.devcontainer/entrypoint.sh:553` and the `installArgv` tail in `.oh/cli/src/lib/tools/catalog.ts` both pre-create `/var/run/tailscale` as `sandbox:sandbox 0755`. `grep -rn 'socket=' docs/ .oh/skills/t3/` → empty; `TS_SOCKET` appears nowhere. All three daemon commands and `tailscale up` are now bare.
- **The pre-creation is genuinely required, not belt-and-braces.** `cmd/tailscaled/tailscaled.go` `MkdirAll`s only the *state* dir; `safesocket`'s Linux `listen()` does `os.Remove(path)` and binds, and never creates the socket's parent. A non-root daemon cannot create `/var/run/tailscale` itself, so without this line the fix would not work.
- **The default path matches.** `paths.DefaultTailscaledSocket()` returns `/var/run/tailscale/tailscaled.sock` whenever `/var/run` stats as a directory. Confirmed in the image: `/var/run -> /run`, which is a real directory.
- **Permissions are sufficient.** `0755` owned by `sandbox` gives the daemon the directory write bit it needs to unlink and bind, and gives the same-user CLI traverse plus socket access. Daemon and CLI are both `sandbox`, so socket ownership is never the constraint.
- **No wipe window.** `/run` is overlayfs here, not tmpfs, and the block runs on every container start under the same `INSTALL_TAILSCALE` gate — so it is recreated on `oh restart` / recreate regardless. The daemon is only ever started by hand afterwards, so nothing races it.
- **The `2>/dev/null || true` asymmetry is safe.** The entrypoint suppresses so a failure cannot break boot; the catalog runs under `set -e` with this as the final command, as root, on a path that exists — practically unfailable, and a failure would be a loud non-zero from `oh tool install` *after* the binaries are already in place, which an idempotent re-run clears. Parity test re-run: `tool-catalog.test.ts` 30/30 green.

### SHOULD-FIX 1 (Node floor) — FIXED, and it is now a real oracle

The `grep '22.16'` is gone, replaced by a stubbed `node v22.15.0` run of `doctor`. I re-broke it two ways:
- `elif ! node_version_ok` → `elif false`: **REGRESSION** ("accepted Node v22.15.0 — the Node floor is not enforced at runtime")
- keep the check but drop `${NODE_RANGE}` from the error string: **REGRESSION** ("does not name the supported range")

Restore → PASS. Both the offending version and the range are independently asserted; neither can be satisfied by text sitting in the help block.

### SHOULD-FIX 2 (log mode) — FIXED, and the unrequested `chmod` is the load-bearing half

Verified empirically: on a pre-existing `0644` log, `(umask 077; : > "$LOG")` alone leaves it `0644` — truncation does not change mode — and the `chmod 600` takes it to `0600`. A subsequent `tee` preserves `0600`. So the extra line the executor added unprompted is the part that actually closes the case it was aimed at. No other writer touches the file.

### Delta hygiene

No comments added to tracked code. No compose, devcontainer, capability, device, privileged flag or published port change — `tailscale-tool-boundary.sh` still PASSes. The PR body's "Not verified here" now names the exact gap the blocker hid in, including that the fix rests on inspection of upstream's default socket path rather than a live round trip. That is accurate and complete.

### One new NIT

`/var/run/tailscale` is the only step of the install recipe duplicated across `entrypoint.sh` and `catalog.ts` with **no drift guard** — `grep -rn 'var/run/tailscale' .oh/cli/src/__tests__/ .oh/evals/probes/` is empty, while version, both shas, the download base and the `/usr/local/bin` paths are all parity-asserted. The one line that fixes the blocker is the one line that can silently drift.

Minimal fix, in the `tailscale matches the entrypoint` block of `tool-catalog.test.ts`:

```ts
it("pre-creates the default tailscaled socket directory in both places", () => {
  expect(ts.installArgv!.join(" ")).toContain("/var/run/tailscale");
  expect(ENTRYPOINT).toContain("/var/run/tailscale");
});
```

### Acceptance criteria movement

| # | Before | Now |
|---|---|---|
| 3 | NOT MET | **MET-BY-DOCS-ONLY** — the recipe is internally consistent and matches upstream's default socket; a live tailnet join remains unexecutable here and is disclosed |
| 4 | NOT MET | **MET** — the preflight no longer false-positives, and the launch string `npx --yes t3 serve --tailscale-serve` under detached tmux is proven by an executing probe. T3's own emission of the `.ts.net` pairing URL stays upstream-dependent and disclosed |
| 9 | partial | **MET** — missing binary, unreachable daemon, and out-of-range Node are all exercised by stubbed runs asserting both exit code and message content |

Carried forward as optional, none blocking: the boundary probe still does not guard `network_mode: host` / `sysctls:` / `devcontainer.json` `appPort`; `tailscaled_hint()` still shells out to `cat`; `docs/connecting.md` step 2 still does not tee `/tmp/agent-tailscaled.log` while the reference and its session table do.

ryaneggz:
## Audit complete — sign-off

`0c3b2b58` verified. `git diff 5a3be14b..0c3b2b58` is exactly one file, `+5 −0`: the parity guard in the `tailscale matches the entrypoint` block. Nothing else in the tree moved.

**The guard is load-bearing, not tautological.** Break-tested in both directions:

- catalog path drifted to `/var/run/tsX` → `1 failed | 30 passed`, and the one failure is `pre-creates the socket directory the entrypoint also pre-creates`
- the `install -d … /var/run/tailscale` line deleted from `entrypoint.sh` → same single failure

Restored → **31/31 pass**, no collateral either way. Both new probes still PASS, and the working tree matches `origin` exactly.

### Findings ledger — all closed

| Finding | Status |
|---|---|
| BLOCKER 1 — custom `tailscaled` socket made the documented flow non-functional | Closed in `5a3be14b`; mechanism verified against upstream `paths`/`safesocket`/`tailscaled` source, not just against the diff |
| SHOULD-FIX 1 — Node-floor assertion was Goodhartable | Closed in `5a3be14b`; re-broken two independent ways, both REGRESS |
| SHOULD-FIX 2 — pairing token in a 0644 log | Closed in `5a3be14b`; mode transition verified empirically, `tee` preserves `0600` |
| NIT — the one install step with no drift guard | Closed in `0c3b2b58`; break-tested above |

CI green 4/4. **Merge-ready.**

### Standing caveat — manual acceptance still required

This is unchanged and correctly disclosed in the PR body. No `tailscale` binary was ever executed in this environment, and there is no tailnet and no phone here. These remain documented-only and want a human pass on a real tailnet before the feature is called proven end to end:

- a real tailnet join, and `t3 serve --tailscale-serve` actually publishing on 443 and printing a `.ts.net` URL
- `t3 pair --tailscale` minting a token against a live server, and the `url` regex matching T3's real output
- phone pairing and reconnect in the T3 Code mobile app
- `BackendState` parsing against real `tailscale status --json`

"A phone outside the tailnet cannot reach the backend" stays discharged structurally — no published port, no capability, loopback-only — by a probe I mutation-tested rather than by a live negative test.

Three optional NITs remain open and none blocks merge: the boundary probe does not guard `network_mode: host` / `sysctls:` / `devcontainer.json` `appPort`; `tailscaled_hint()` shells out to `cat`; `docs/connecting.md` step 2 does not tee `/tmp/agent-tailscaled.log` while the reference and its session table do.

Issue #898, opened 2026-08-30
task(#898): collapse sandbox storage to one $HOME mount


ryaneggz:
## Problem

The sandbox declares **11 named volumes** — one per agent tool (`claude-auth`, `codex-auth`, `pi-auth`, `opencode-auth`, `grok-auth`, `deepagents-auth`, `herdr-data`, `cloudflared-auth`, `ssh-config`, `config-dir`, `cc-safety-net`) — plus `oh_workspace` in the image-only flavor.

Every new tool means a new volume, a new mount line, a new entry in `entrypoint.sh`'s hardcoded chown list, and a new line in the docs. State written anywhere else under `$HOME` (shell history, `~/.aws`, a tool nobody anticipated) is silently lost on recreate.

## Scope

1. **One mount** at `/home/sandbox` replaces all 11 volumes.
2. **`oh init` asks for a host path** (`storage.homePath` → `OH_HOME_MOUNT`). Blank means the Docker-managed `<sandbox-name>_workspace` volume.
3. **Retire the `OH_PROJECT_ROOT` config knob** — the project root is already nested inside the one mount at `/home/sandbox/harness`, so a configurable location for it is redundant.

## Load-bearing detail

An empty named volume gets the image's `/home/sandbox` auto-copied in; a host bind does **not**, and would shadow `.oh-my-zsh`, `.zshrc`, and `/home/sandbox/.local` (`NPM_USER_PREFIX`). The fix is to empty the image's home into `/opt/home-seed` at build time so a single entrypoint seed populates either mount kind identically.

Separately, `entrypoint.sh`'s `find /home/sandbox -xdev -uid ...` only prunes the repo bind today because `/home/sandbox` is the container overlay. Once it is a mount, `-xdev` stops pruning and the whole checkout would be chowned on every UID change. It must become an explicit `-path "$OH_PROJECT_ROOT" -prune`.

## Breaking

Existing sandboxes lose the 11 volumes. Migration is a documented `docker cp` recipe, not automation.

PR #901, opened 2026-08-30
FROM task/900-slim-sandbox-image TO task/898-single-home-mount

### LINKED ISSUE #900: task(#900): slim the sandbox image now that $HOME persists
Follow-up to #898 / #899.

## Idea

Once the whole sandbox home persists through the single `/home/sandbox` mount, the image no longer has to carry what the home can hold. The Dockerfile could shrink to roughly "a base runtime plus the `oh` CLI", and provisioning would move inside the running sandbox where its results persist in the mount.

## Why this is now possible

Before #898, anything installed into `$HOME` at runtime was lost on recreate unless a volume happened to cover its directory — so tool installs had to be baked into the image. That constraint is gone: `~/.local` (`NPM_USER_PREFIX`), `~/.oh-my-zsh`, `~/.cache`, and every agent CLI's state now survive in the one mount.

`oh harness <list|install|status>` (#821) already installs optional harnesses into a running sandbox without a rebuild, so part of the machinery exists.

## Sketch

- Dockerfile keeps: base image, apt runtime deps, the `sandbox` user, the `oh` CLI, the entrypoint.
- Dockerfile drops (moves to first-boot / on-demand provisioning): the agent CLIs, oh-my-zsh and its plugins, the uv/python tree, herdr.
- `/opt/home-seed` shrinks to the dotfile skeleton, which also cuts the ~200 MB that #899's seed layer costs today.
- First boot provisions into the mount; later boots find it already there.

## Open questions

- Cold-start cost: a fresh sandbox trades image pull time for provisioning time. Worth measuring both before committing.
- Reproducibility: pinned versions currently live in the Dockerfile and are enforced by probes (`cc-safety-net-wiring.sh` checks the image pin; `herdr-default.test.ts` checks the herdr sha256). Runtime provisioning needs an equally checkable pin surface.
- Offline/air-gapped boots stop working if provisioning requires the network at first boot.
- CI's boot guard builds and boots the image; the smoke test's expectations would move from build time to boot time.

Not started — filed so the idea is not lost.

ryaneggz:
Closes #900.

Stacked on #899 (`task/898-single-home-mount`) — Phase 2 restructures the seed staging that #899 introduced. GitHub will retarget this to `development` when #899 merges.

## #900's premise did not survive measurement

The issue proposed shrinking the Dockerfile to "base runtime + `oh` CLI" and moving the agent CLIs, oh-my-zsh, the uv/python tree, and herdr to runtime provisioning, on the theory that the single `/home/sandbox` mount now makes runtime installs persist.

It does not, for the artifacts that matter. The three largest — `@openai/codex` (336 MB), `@anthropic-ai/claude-code` (239 MB), `cc-safety-net` (25 MB) — install to `/usr/lib/node_modules`, **outside** the mount, because the root npm prefix is `/usr`. Moving them to `oh harness install` would make them re-download on **every container recreate** — a regression in the exact property #898 was built to establish. Making them persist first requires relocating the root global prefix into `$HOME`, which *grows* the seed by 600 MB unless you also stop baking them, at which point first boot needs a ~600 MB network install before `claude` exists.

Two further costs: `.oh/cli/src/lib/harnesses/catalog.ts` installs **unpinned** (`npm install -g @anthropic-ai/claude-code`, no version), so runtime provisioning means inventing a pin-and-verify surface that build time already has for free; and `verify-sandbox-image.sh` would stop being an image oracle and become a booted-container oracle, network-dependent inside the boot window.

So this PR takes the reachable bytes instead — the 358 MB baked home — and takes them by deletion rather than by moving work to boot time. No install moves to runtime, no new pin surface, offline first boot still works, and the prebuilt-image "comes up in the time it takes to pull" promise is intact.

Full reasoning, including the rejected middle path, is in `.oh/tasks/slim-sandbox-image/prd.md`, committed here.

## What changed

**Phase 1 — stop shipping build caches.** The seed measured 358 MB, of which 103 MB was pure build cache: `~/.npm` (31 MB, left by the pi npm install) and `~/.cache/uv` (72 MB, left by `provision-python.sh`). Neither is read at runtime — `entrypoint.sh`'s `repair_home_mount_ownership` `install -d`s the uv cache on every boot and `provision-python.sh` does so again, while npm recreates its own on first use. Both purges fold into the `RUN` that produces the cache, rather than sitting in a late `rm -rf` layer, so they travel with their producer into the builder stage.

**Phase 2 — store the seed once.** `RUN mv /home/sandbox /opt/home-seed` moved data across a layer boundary: the new layer holds a full copy while every earlier layer still holds the original, so the image stored the seed twice. The Dockerfile is now three stages — `base` (through the `sandbox` user and the `UV_*` env), `home` (everything that writes into `/home/sandbox`), and `final`, which brings the seed in via `COPY --from=home --chown=sandbox:sandbox`.

**`.dockerignore`** now excludes `.pnpm-store` and the untracked `.pi` build outputs, which were pulling ~2 GB into every local build context. CI builds from a clean checkout and was never affected; every operator's local `oh sandbox` build was.

## Measured

`docker system df -v`, both images built from this checkout:

| image | total | unique layers |
|---|---|---|
| baseline | 4.0 GB | 1.981 GB |
| this PR | **3.46 GB** | 1.442 GB |

**~540 MB.** The seed itself: baseline `/opt/home-seed` 358 MB (`.npm` 31 MB, `.cache/uv` 72 MB); after, 319 MB — the identical payload, stored once instead of twice.

The baseline build kept the new `.dockerignore` deliberately. With the old one, `COPY . /opt/oh-seed/` would have swallowed this dev sandbox's 1.5 GB `.pnpm-store` and 509 MB `.pi` — artifacts absent in CI — and inflated the delta dishonestly. The 540 MB is the Dockerfile change alone.

## Two non-obvious things

**`rm -rf /home/sandbox` is required in the final stage.** The old wholesale `mv` took the `useradd -m` skel files (`.bashrc`, `.profile`, `.bash_logout`) with it. A bare `install -d` would have left them behind, breaking the empty-home invariant that stops an empty named volume from auto-copying — the load-bearing property from #899.

**`COPY . /opt/oh-seed/` is duplicated** in `home` and `final`, at the same relative position as before. `provision-python.sh` reads it from the home stage; the final image needs it for `seed_workspace_volume`. Only the final stage's copy is stored. Hoisting it into `base` would make every repo edit invalidate the pi install, oh-my-zsh, and the Python kernel.

Two `ARG`s move with their use (`INSTALL_PYTHON_KERNEL`, `OH_PYTHON_VERSION` into `home`) and `AGENTS` is re-declared there, since an `ARG` is stage-scoped. All eight were audited; the optional-installer `ARG`s stay wholly within `base`.

## Oracles

Two moved with the mechanism, both deliberately:

- `oh-home-mount.sh:55` pinned the literal `RUN mv /home/sandbox /opt/home-seed` **spelling**. It now asserts the invariant — seed staged at `/opt/home-seed`, `/home/sandbox` empty, `harness/` absent — satisfied by either mechanism.
- `sandbox-base-image.test.ts:11` end-anchored `/^FROM node:22-trixie-slim$/m`, which a stage alias necessarily breaks (`COPY --from` cannot reference a stage by index). Relaxed to allow ` AS <alias>` while keeping the pin.

New tier-A probe `image-seed-hygiene.sh` asserts the caches are purged **no later than** the seed-staging layer and that `.dockerignore` excludes the store directories. It was written against the invariant rather than a spelling and verified to survive the Phase 2 restructure before that restructure was written. Negative controls: removing either purge, removing the `.dockerignore` lines, or moving a purge *after* the staging line each drives it to exit 1.

## Verification

- `verify-sandbox-image.sh` against the built image: **exit 0**, all 14 checks (trixie base, apt suites, 1000:1000, node 22.23.2, pnpm 10.33.0, herdr 0.7.4 + checksum, tool versions).
- Eval suite: **exit 0**, 102 probes, no green→red.
- Root `npm test`: 930 passed, 8 failed — all in `compose-args.test.ts`, verified pre-existing via a stashed baseline (this dev sandbox has SSH-overlay conditions the fixtures assume absent).
- Image inspection: `/opt/home-seed` is `sandbox:sandbox`, zero non-sandbox-owned entries, `.ssh` mode `0700` preserved through `COPY --from`; `.npm`/`.cache/uv`/`harness` absent; `/home/sandbox` holds only the `WORKDIR`-created `harness/`, as before.
- Booted container, fresh volume: `seed_home` populated the mount, `provision-python` verified Python 3.11.16 + ipykernel, providers linked, and `pi`/`claude`/`codex`/`herdr`/`uv` all resolve.

Bare `python` is not on `PATH` — confirmed identical in the baseline image, not a regression. The kernel interpreter resolves the documented way via `PRIME_AGENT_KERNEL_PYTHON`.

## Not done here

Flipping `INSTALL_PYTHON_KERNEL` to `false` would cut another 166 MB, but it is the first change that makes first boot depend on the network, and it is a policy decision about default capability rather than image hygiene. Worth a separate issue if image size is still a pain.


ryaneggz:
## Adversarial audit round (commit `a3e05728`)

`/audit pr` returned **PR-AUDIT-PROMOTABLE**, but that route grades promotability evidence and says outright that diff correctness is outside its scope. A separate adversarial audit of the diff found four defects that all six green checks missed. All are fixed here.

**The restructure itself held up.** The highest-value check was an entry-by-entry comparison of `/opt/home-seed` between the baseline and multistage images: **29,992 common entries, zero differences in type, mode, or owner.** Symlinks stayed symlinks, `.ssh` stayed `drwx------`, `install/*.sh` exec bits intact. The path-set delta is exactly the intended purge — 8,584 entries under `.cache/uv`, 833 under `.npm`, nothing else, and zero paths present only in multistage.

The defects were all in the guards.

**D1 — the new probe passed a Dockerfile that still shipped both caches.** `image-seed-hygiene.sh` compared textual line numbers in a file where instruction order is not execution order. The accepted mutant: delete both purges from `home`, add `RUN rm -rf /home/sandbox/.npm /home/sandbox/.cache/uv` as the first instruction of `FROM base AS final`. It is textually before the `COPY --from=home`, so the line check passes — while it purges `final`'s throwaway home and the seed is copied from an unpurged stage. This is the exact failure the probe existed to prevent, and it was the probe's only substantive assertion.

It now parses stages, anchors on the staging instruction rather than any mention of `/opt/home-seed`, resolves which stage the seed is copied *from*, and requires a purge inside that stage. The single-stage `RUN mv` fallback still works. The old anchor also false-positived on any earlier `ENV` referencing the path; that is fixed and has its own control.

**D2 — `sandbox-base-image.test.ts` stopped pinning anything.** With three `FROM` lines, any one matching satisfied the regex; `FROM ubuntu:24.04 AS final` passed all five tests. A new case walks the `FROM` lines in order and asserts every image is either the pinned base or a stage defined earlier.

**D3 — `oh-home-mount.sh` accepted sources that establish nothing.** It took `COPY --from=base` (whose home is bare `useradd -m` skel, so every baked dotfile would be silently lost) and `COPY --from=alpine:latest`. The source must now be a stage declared in the same Dockerfile, and not `base`.

**D4 — `/opt/home-seed` silently went 0700 → 0755.** `RUN mv` carried the directory's own mode; `COPY --from` creates the destination at the default 0755 and `--chown` does not restore a mode. Low impact — no secrets in the seed, `.ssh` inside is still 0700, `seed_home` runs as root — but it is an unintended relaxation of exactly the class the #899 audit caught, and no oracle covered it. Restored, and now asserted.

**Mutation-verified.** The audit's mutant and all four original negative controls exit 1; the false-positive control stays green; both bad `COPY --from` sources are rejected by both probes; the D2 mutant fails the test suite. The single-stage fallback was exercised separately in a synthetic tree.

### Known gaps

- The `optional-installers` path (`INSTALL_HERMES`/`DEEPAGENTS`/`OPENCODE`/`GROK_BUILD` on) was not built locally, so `final`'s `chown -R /opt/uv` and the hermes chown were never exercised against populated trees. CI builds that image, but `verify-sandbox-image.sh` inspects neither `/opt/uv` ownership nor `/opt/home-seed`, so no oracle covers it.
- Local builds used `DOCKER_BUILDKIT=0` and amd64 only.
- D4's `chmod` was not re-verified against a rebuild; the new probe assertion is what holds it.

PR #903, opened 2026-08-31
FROM task/902-provision-harnesses TO development

### LINKED ISSUE #902: task(#902): provision default harnesses into the home mount instead of the image
Follow-up to #900, first of three.

`oh` is meant to split by execution target: on the host it provisions the host or the sandbox; **inside the sandbox it provisions the sandbox with harnesses and tools** (`docs/lifecycle-commands.md` § "Where you are standing when you type `oh`").

Two things break that today.

## `oh harness install` does not work inside the sandbox for the default harnesses

`claude-code` and `codex` carry `installUser: "root"` in `.oh/cli/src/lib/harnesses/catalog.ts`. Against the local execution target that becomes `sudo -n -- npm install -g …` (`.oh/cli/src/lib/execution/local-target.ts:114-118`), and `/etc/sudoers.d/sandbox` grants `sandbox ALL=(ALL) ALL` with **no `NOPASSWD`**. `sudo -n true` inside the sandbox returns `sudo: a password is required`, rc=1. So the documented in-sandbox half of the split fails for the two harnesses most people use.

`pi` already does it correctly: `installUser: "sandbox"` with `npm --prefix /home/sandbox/.local install -g`. It installs into the home mount, so it also persists across container recreate.

## Baked installs are unreachable by the CLI

Anything at `/usr/lib/node_modules` can only change by rebuilding the image. Anything under `/home/sandbox/.local` can be upgraded in place by the CLI in a running remote sandbox. Since #898 made the whole home persist, the second is now strictly better for anything the catalog knows how to install.

## Scope of this issue

Add the provisioning mechanism and fix the catalog. **Remove nothing from the image** — the size win is #900's follow-up, and keeping the bake makes this independently revertible.

- New `.oh/scripts/provision-harnesses.sh`, modelled on `provision-python.sh` (same mode flags, same root→`gosu sandbox` re-exec, same failure messages).
- `claude-code` and `codex` move to `installUser: "sandbox"` + `npm --prefix /home/sandbox/.local install -g`. **`claude-code` must not use `--ignore-scripts`** — its `postinstall` copies the native binary over a placeholder, and without it `claude --version` fails with `Error: claude native binary not installed`. Verified both ways.
- `entrypoint.sh` gains `OH_PROVISION_HARNESSES` (default true), mirroring `OH_PROVISION_PYTHON`. Provisioning **warns, it does not fail the boot** — an offline sandbox must still come up as a usable shell.
- `.devcontainer/Dockerfile` gains `ARG BAKE_HARNESSES=true` gating the existing agent-CLI block. Default stays `true`, so this PR changes no shipped image content.
- A probe asserting no `default`-kind harness uses `installUser: "root"` and that each installs into `NPM_USER_PREFIX`.

`cc-safety-net` stays baked — operator ruling. `link-providers.sh:158` hard-fails the boot when it is absent under `CC_SAFETY_NET_STRICT=1`, and it runs before any provisioning could happen.

ryaneggz:
Closes #902. First of three following #900; nothing leaves the image here.

## The bug this fixes

`oh` splits by execution target: on the host it provisions the host or the sandbox, and **inside the sandbox it provisions the sandbox with harnesses and tools** (`docs/lifecycle-commands.md`). The second half did not work for the two harnesses most people use.

`claude-code` and `codex` carried `installUser: "root"`. Against the local execution target that becomes `sudo -n -- npm install -g …` (`.oh/cli/src/lib/execution/local-target.ts:114-118`), and `/etc/sudoers.d/sandbox` grants `sandbox ALL=(ALL) ALL` with **no `NOPASSWD`** — `sudo -n true` inside the sandbox returns `sudo: a password is required`, rc=1.

Both now match the `pi` entry directly above them: `installUser: "sandbox"`, `npm --prefix /home/sandbox/.local install -g`. That also lands them **inside the home mount**, so since #898 they survive container recreate and can be upgraded in place in a running remote sandbox instead of requiring an image rebuild. Pi has worked this way all along — it is the existence proof, not a new idea.

**`claude-code` deliberately does not get `--ignore-scripts`.** Its `postinstall` (`install.cjs`) copies the native binary over the `bin/claude.exe` placeholder. Verified both ways against a scratch prefix: with scripts, `claude --version` → `2.1.251`; with `--ignore-scripts`, `Error: claude native binary not installed`. The probe asserts the flag stays off, because `pi` next door uses it and it would be an easy "consistency" edit to make.

## The provisioner

`.oh/scripts/provision-harnesses.sh` follows `provision-python.sh` — same mode flag, same root → `gosu sandbox` re-exec, same `check_writable` ownership diagnostics, same die-with-the-exact-command style. `--print-env` is absent because this provisioner exports nothing downstream, unlike the Python one.

**It reads the catalog through the CLI**, not by re-encoding it: `oh harness list --json` (which already runs each entry's `verifyArgv` to compute `installed`), then `oh harness install <id> --no-persist` for each missing one. The shell knows no ids, no packages, no prefixes, no argv — the TypeScript catalog stays the only description of what a harness is, per AGENTS.md's rule against a second unverified description.

**No `kind: "default"` harness carries a version pin today**, so an existing install is never replaced. The script says exactly that in its output (`OK claude-code present (unpinned — an existing install is never replaced)`) rather than implying it refreshes. When pins arrive they belong in the CLI, where every caller gets them, not in a shell version-compare only the boot path would use.

Two hardening details: it `cd`s to the repo root derived from `${BASH_SOURCE[0]}` (the CLI's `resolveProjectRoot` walks up from cwd, which the entrypoint does not guarantee), and it exports `OH_EXECUTION_TARGET=local` (`runningInsideSandbox` needs a non-empty `SANDBOX_NAME`, which is not reliably set at entrypoint time — without this the CLI would pick the docker-compose target from inside the container).

## Boot ordering

The hook runs **after** `link-providers.sh --init`, not before. `link-providers`'s only binary dependency is `cc-safety-net` (`:161-177`), which stays baked, and its symlink work never touches `claude`/`codex`/`pi`. It is also the boot-critical hard gate (`exit 1`), so a network-dependent best-effort step does not belong in front of the step that decides whether the boot is viable.

Provisioning **warns and continues**. An offline sandbox still comes up as a usable shell — zsh, git, `gh`, herdr, the checkout and the `oh` CLI are all baked. The image-only flavor runs `restart: unless-stopped`, so a hard-failing provisioner would crash-loop.

## `BAKE_HARNESSES`

Defaults to `true`, so **this PR changes no shipped image content**. It gates only the `$AGENTS` loop, not the whole `RUN` — `INSTALL_OPENCODE` and `INSTALL_GROK_BUILD` are separate opt-ins and turning them off as a side effect would be a silent regression. Flipping the default and dropping the bake is the next PR, so a boot regression bisects to one commit.

`cc-safety-net` is untouched and stays baked unconditionally, per operator ruling — `link-providers.sh:158` hard-fails the boot without it under `CC_SAFETY_NET_STRICT=1`, and that check runs before anything could provision it.

## Probe

New `harness-home-provisioning.sh` rather than an extension of `tool-catalog-boundary.sh`: that probe's contract is catalog disjointness plus agent-browser's download gate, a different lesson with a different `# source:`, and folding this in would make a green→red there ambiguous about which invariant broke.

Nine negative controls, each applied and reverted — `installUser` back to root (1), prefix dropped (1), `--ignore-scripts` added (1), env var renamed (1), entrypoint no longer naming the script (1), warn-and-continue text removed (1), `ARG` renamed (1), script `chmod -x` (1), and `ENV NPM_USER_PREFIX` removed → **exit 2, SKIPPED**, correctly reporting unverifiable rather than passing.

## Verification

| | |
|---|---|
| `bash .claude/skills/eval/run.sh` | exit 0 — new probe PASS; `oh-config-surfaces` REGRESSION pre-existing and unchanged |
| `npm test` (repo root) | 938 passed, 8 failed — all `compose-args.test.ts`, the known environmental baseline; `harness-catalog.test.ts` 33/33 |
| `cd .oh/cli && npm run build` / `npm run typecheck` | 0 / 0 |
| CI's shellcheck invocation | 0 |

**Exercised for real:** the provisioner end-to-end twice in a live sandbox (idempotent; warm run 3.3 s, dominated by the nine `verifyArgv` execs behind one `harness list --json`); `--verify` against a scratch prefix with a stripped `PATH`, correctly reporting `pi` absent and exiting 1; the usage error (2); and the stale-CLI path. Both install argv proven on a scratch prefix; the live `claude`/`codex` binaries were never touched.

**Reasoned about, not run:** the Dockerfile change (no image build) and cold-boot ordering in a fresh container. CI covers both.

## One thing to know

The baked `oh` in an **already-running** container can predate `oh harness` — this dev container's does (v0.1.1). The provisioner detects that and dies loudly-but-non-fatally with "rebuild the sandbox image", and the boot still comes up. A released image builds `/opt/oh` from the same commit, so this only affects containers straddling the change.


ryaneggz:
## Adversarial audit round (`d91cda31`)

Six green checks did not catch five defects. The most serious is a boot hang.

**D1 — the boot path could hang indefinitely.** `oh harness list --json` probes **every** entry in the catalog, not just the three defaults. One is `t3code`, whose `verifyArgv` is `npx --no-install t3 --version` — `npx` contacts the registry, and `probeInstalled` passed no `timeoutMs`, so `spawnSync` waited without bound. Reproduced against an unreachable registry: still running at **2m30s** when the auditor's own `timeout` killed it.

On any boot where DNS resolves but the registry does not answer — egress-restricted VM, proxy outage, registry incident — the entrypoint blocks before `sleep infinity`, exceeds the 300 s `start_period`, and never goes healthy. `restart: unless-stopped` does not rescue an unhealthy-but-alive container, and warn-and-continue cannot help because a hang never reaches the `if !`. It hangs while *listing*, before any install, so the shipped `BAKE_HARNESSES=true` did not avoid it.

Bounded at three layers, because each fails differently:

| layer | |
|---|---|
| `timeoutMs: 15000` on the probe spawn | a hung `verifyArgv` cannot stall the CLI; a timeout maps to *unknown*, never a crash |
| `oh harness list --defaults` | the boot path probes 3 entries instead of 9 and never runs `npx` |
| `timeout` on the entrypoint call | the boot is bounded whatever the CLI does |

Measured against the auditor's exact command: **2m30s (killed) → 17.0 s full-catalog, 1.55 s with `--defaults`.**

**D2 — the install loop silently skipped harnesses and reported success.** It read from a herestring while installs run with `stdio: "inherit"`, so an installer that reads stdin consumed the rest of the loop. Reproduced with a stub: three missing harnesses, **one installed, exit 0, "OK default harnesses provisioned" printed.** Latent with npm, live the moment a default uses the `curl … | bash` shape that `hermes` and `prime-agent` already use. Installs now read from `/dev/null`.

**D3 — the remediation command provisioned the host.** The script force-exported `OH_EXECUTION_TARGET=local`, which short-circuits the in-container check, while the prefix is hardcoded to `/home/sandbox/.local` and every error said "run: bash .oh/scripts/provision-harnesses.sh" with no mention of where. It now refuses outside the sandbox, reusing the CLI's own `runningInsideSandbox` predicate rather than inventing a check.

Found while fixing it: the documented raw `docker run` recipe (`docs/deployment-prebuilt-image.md:215-221`) never passes `SANDBOX_NAME`, so the new guard would have silently skipped provisioning for the prebuilt-image flavor. The entrypoint now asserts the local target explicitly, which keeps the host guard intact for manual invocation.

**D4 — the probe passed a Dockerfile whose gate had been deleted.** It asserted `ARG BAKE_HARNESSES=true` was *declared*, not that anything referenced it. Replacing the `if` with an unconditional `{` left it green. It now checks the ARG is referenced by the `RUN` that installs `$AGENTS` **and** by the one that bakes pi, and that both stages declare it — deleting either declaration also used to pass.

This is the third probe in this series to pass the exact thing it was written to reject. Every assertion here was broken and confirmed to exit 1 before being restored.

**D5 — `BAKE_HARNESSES=false` did not unbake pi.** `ARG` is stage-scoped; the flag was declared in `base` but the `home` stage bakes pi gated only on `AGENTS`, while the else-branch message claimed the provisioner installs "them" at boot. Harmless today, actively misleading in PR-2, which flips this default. The `home` stage now declares and honors the flag.

Also removed the illusory `OH_SANDBOX_USER` knob (the catalog hardcodes both the user and the prefix, so a non-default value failed with a message that blamed the wrong thing), and the final log line no longer claims to have provisioned anything in `--verify` mode.

### Left open

- **L3** — the provisioner probes via `harness list`, then `harness install` re-probes per id. `--defaults` cut this from 9+N to 3+N; the duplication remains.
- **L4** — an image whose baked `oh` predates `oh harness` (or now `--defaults`) emits a `die`-then-warn on every boot. Same remedy as before ("rebuild the sandbox image"), no new failure mode, but the window widened by one release.
- **Not exercised anywhere:** with `BAKE_HARNESSES=true` shipping as the default, the *install* path never runs at boot. D1–D3 all live in code that CI's green run and a normal boot both skip. PR-2 is what will exercise it, which is an argument for landing PR-2 behind a real boot test rather than trusting this PR's green.

PR #905, opened 2026-08-31
FROM task/904-unbake-harnesses TO development

### LINKED ISSUE #904: task(#904): stop baking default harnesses into the sandbox image
## Problem

PR #903 added `.oh/scripts/provision-harnesses.sh` and wired it into the boot path, but
shipped it behind `ARG BAKE_HARNESSES=true`. With the bake still on, every default
harness is already present when the provisioner runs, so **the install path never
executes at boot**. Four defects found in the #903 audit all lived in code that a green
CI run and a normal boot both skip.

The image also still owns harness provisioning, which contradicts the boundary:
outside the sandbox the CLI provisions the host and the sandbox; **inside** the sandbox
the CLI provisions harnesses and tools. `@anthropic-ai/claude-code`, `@openai/codex`,
and `@earendil-works/pi-coding-agent` are `kind: "default"` harnesses — the in-sandbox
CLI's responsibility, not the Dockerfile's.

## Change

- Delete the bake. Remove `ARG BAKE_HARNESSES`, `ARG AGENTS`, the `PKG` map, the
  `$AGENTS` loop in `base`, and the gated pi install in `home`. No dormant alternative.
- Raise the healthcheck `start_period` so boot-time provisioning fits inside it.
- Give the boot smoke a real post-boot oracle: assert every default harness is present
  under `/home/sandbox/.local`, owned by `sandbox`, after a fresh-volume boot.
- Add a negative image oracle: `verify-sandbox-image.sh` fails if the image ships a
  default harness.
- Invert the probe and unit assertions from "the bake is gated" to "there is no bake".

`cc-safety-net` stays baked (operator ruling). Opt-in `INSTALL_*` harnesses
(opencode, grok-build, deepagents, hermes) are unaffected.

## Cost this accepts

- First boot on a fresh home volume now needs network and takes ~60–180s longer.
- An offline first boot yields a usable shell with no agent CLIs (warn-and-continue).
- `~/.npm` moves into the home volume and grows across upgrades.

ryaneggz:
Closes #904

**Stacked on #903** (`task/902-provision-harnesses`). Retarget to `development` after that merges.

## Why

#903 added `.oh/scripts/provision-harnesses.sh` and wired it into the boot path, then shipped it behind `ARG BAKE_HARNESSES=true`. With the bake on, every default harness is already present when the provisioner runs — so **the install path never executes**. All four defects that PR's audit found lived in code that a green CI run and a normal boot both skip. This PR is what makes that code load-bearing, and it does not land without an oracle that runs it.

It also finishes the ownership boundary: outside the sandbox the CLI provisions the host and the sandbox; **inside** the sandbox the CLI provisions harnesses and tools. `@anthropic-ai/claude-code`, `@openai/codex`, and `@earendil-works/pi-coding-agent` are `kind: "default"` harnesses — the in-sandbox CLI's job, not the Dockerfile's.

## What changed

**The bake is deleted, not disabled.** `ARG BAKE_HARNESSES`, `ARG AGENTS`, the `PKG` map, the `$AGENTS` loop in `base`, and the gated pi install in `home` are gone. A build arg that can re-bake is a dormant path that would restore both the shadowed `/usr/lib/node_modules` copy (root-owned, unwritable from a running sandbox, so `oh harness install` cannot upgrade it) and the untested boot install. The probe rejects the ARG's return.

**The install path is now exercised and asserted.**

| Oracle | Assertion |
|---|---|
| `sandbox-boot-smoke.sh` | Each default harness resolves under `NPM_USER_PREFIX` via `type -P`, is owned by the *reconciled* sandbox uid, and prints its own version. Refuses to pass when the catalog reports no defaults. |
| `verify-sandbox-image.sh` | Reads the catalog out of the image itself; no `kind:"default"` harness may be installed. Fails loudly if the catalog is unreadable. |
| `sandbox-boot-guard-ci.sh` | Derives the smoke deadline from the compose healthcheck window (`start_period + interval × retries`) instead of pinning a literal. Also rejects `OH_PROVISION_HARNESSES: false` in the guard. |
| `harness-home-provisioning.sh` | Inverted from "the bake is gated" to "no default harness package appears in the Dockerfile", reading package names out of `installArgv` so it cannot drift from the catalog. |

The boot guard tears down with `down -v`, so its smoke boots on a **fresh home volume** and does real npm work every run. `start_period` goes 300s → 600s in both compose files; `BOOT_SMOKE_TIMEOUT_SECONDS` 900 → 1200 to clear the resulting 690s unhealthy deadline.

`oh harness list --json` gains a `binary` field so an external oracle can check install location without hardcoding it.

## Scope

`cc-safety-net` stays baked (operator ruling). Opt-in `INSTALL_*` harnesses — opencode, grok-build, deepagents, hermes — are untouched; they are root-installed image-level runtimes, not default harnesses.

## Costs this accepts (documented in `docs/installation.md`)

- A first boot on a fresh home mount **needs network** and runs 60–180s longer.
- An offline first boot yields a usable shell with **no agent CLIs** — warn-and-continue, retry with `bash .oh/scripts/provision-harnesses.sh`.
- `~/.npm` moves into the home mount and grows across upgrades (`npm cache clean --force` reclaims it). Today the Dockerfile `rm -rf`s it so the image never ships it.
- An existing install is never replaced, so the provisioner no-ops on every boot after the first.

## Verification

- `EVAL=0` — 103 probes, no regressions. (`oh-config-surfaces` is a pre-existing persistent red, unchanged.)
- `npx vitest run` — 952 passed / 8 failed; the 8 are the known `compose-args.test.ts` environmental baseline, unchanged from `development`.
- shellcheck 0, `npm run build` 0, `tsc --noEmit` 0.
- Local `docker build` of the new Dockerfile — see the follow-up comment.

**Every new assertion was mutation-verified before shipping.** Three probes in this series have passed the exact thing they existed to reject, so each new one was broken deliberately and confirmed to exit 1:

- re-bake claude-code into `base` → caught; re-bake pi into `home` → caught; reintroduce `ARG BAKE_HARNESSES` → caught; a *comment* naming the package → correctly still passes.
- `start_period` raised past the smoke timeout → caught; smoke timeout dropped below the deadline → caught; env var deleted → caught.
- `harnessProbeFails` / `noDefaultHarnesses` / `bakedHarnesses` / `harnessCatalogFails` fixtures cover the boot-smoke and image oracles in CI, not just by hand.

That process caught a real defect in my own check: the first draft used `command -v`, which returns the `.bashrc` alias text rather than a path for `claude` and `codex` in an interactive shell. Switched to `type -P` and invocation by absolute path.

## Not exercised

`OH_IMAGE_ONLY=1` (flavor B) is reasoned about and documented but not booted. amd64 only.


ryaneggz:
## Verification results

### The install path runs, on a fresh volume, in CI

The boot-guard job on this PR is the first run in this series where the provisioner actually does work — the previous PR's green never executed it:

```
Volume "openharness-boot-guard-33349686306_workspace"  Created
sandbox boot smoke: claude-code provisioned at boot -> 2.1.251 (Claude Code)
sandbox boot smoke: codex       provisioned at boot -> codex-cli 0.151.0
sandbox boot smoke: pi          provisioned at boot -> 0.84.4
sandbox boot smoke ok: ... bind-ownership, and boot-provisioned harness checks
```

The image oracle fired in the same run:

```
ok: no default harness is baked into the image (claude-code codex pi)
```

**Boot cost: 21 seconds** (02:09:12 start → 02:09:33 healthy), not the 60–180s I estimated. `start_period: 600s` is therefore generous rather than tight; I have left it, since slower links exist, and corrected `docs/installation.md` to cite the measured figure next to the conservative budget.

### Local image build

`docker build` of the new Dockerfile succeeds. `verify-sandbox-image.sh` passes all 16 checks against the real image, and `type -P` inside it confirms the intended shape:

| Binary | New image |
|---|---|
| `claude`, `codex`, `pi` | absent |
| `cc-safety-net`, `herdr` | present (still baked, as ruled) |
| `/home/sandbox` | one entry — the `harness` mountpoint stub |
| `/opt/home-seed/.local/bin` | `python3.11` only; pi is gone |

### Size

**I cannot give you a clean top-line image-size number, and the obvious one is wrong.**

I built the parent commit for a one-variable comparison and got 3.33 GB → 2.02 GB. That figure is contaminated: the baseline was built from a worktree and this branch from the main checkout, so the two build contexts differ, and `/opt/oh-seed` (the context copy) accounts for roughly 100 MiB of the gap on its own. Reporting −1.31 GB would be overstating the win.

What *is* clean — these three paths are produced by the Dockerfile, not copied from the build context, so the context difference does not touch them:

| Path | Baseline | This PR | Delta |
|---|---|---|---|
| `/usr/local/lib/node_modules` | 579 MiB | 43 MiB | **−536 MiB** (codex 332, claude-code 205) |
| `/root/.npm` | 261 MiB | 20 MiB | **−241 MiB** (their install cache) |
| `/opt/home-seed/.local` | 301 MiB | 166 MiB | **−135 MiB** (pi) |
| | | | **−912 MiB uncompressed** |

Compressed will be materially less — these are already-compressed native binaries. Note the cost side is real and unmeasured here: that 241 MiB npm cache does not disappear, it relocates to `~/.npm` in the operator's home mount and grows across upgrades.

### Local checks

`EVAL=0` (103 probes, no regressions; `oh-config-surfaces` is a pre-existing persistent red). `npx vitest run` 952 passed / 8 failed — the 8 are the known `compose-args.test.ts` environmental baseline. shellcheck 0, `npm run build` 0, `tsc --noEmit` 0.

ryaneggz:
### Retarget note — why this branch carries an `-s ours` merge

#903 landed as a **squash**, collapsing `61b18011` and `d91cda31` into `09dfa8c4`.
This branch already contains both of those as ancestors, so retargeting to
`development` produced a `DIRTY/CONFLICTING` state that was purely history-shape,
not content.

Before resolving I verified that `development` contributes nothing this branch lacks:

- `bee80601` (the squash's parent) **is** an ancestor of this branch.
- Of the 12 files the squash touched, 7 are byte-identical here; the 5 that differ
  (`Dockerfile`, `harness-catalog.test.ts`, `commands/harness.ts`,
  `harness-home-provisioning.sh`, `RESULTS.md`, `CHANGELOG.md`,
  `docs/harnesses/overview.md`) differ only by this PR's own edits on top —
  e.g. the CHANGELOG still carries #902's entry, and `overview.md` evolves
  #903's sentence rather than reverting it.

`git merge -s ours` therefore records the ancestry without reapplying content, and
the post-merge tree is **byte-identical** to `dfde856b`, the pre-merge tip that CI
had already gone green on. Nothing in this PR's reviewed diff changed.

PR #907, opened 2026-08-31
FROM task/906-tools-out-of-image TO development

### LINKED ISSUE #906: task(#906): move herdr and cloudflared out of the image and into the tool catalog
## Problem

Per the operator ruling — *"outside the sandbox the CLI is used for provisioning host or sandbox on the host; once inside the sandbox the CLI is for provisioning sandbox with harnesses and tools"* — herdr and cloudflared are tools, so the in-sandbox CLI owns them. Today both are baked into the image and the tool catalog marks them `kind: "baked-in"` with a `notInstallableReason` telling the operator to rebuild.

#905 did this for the default *harnesses*. This does it for the default *tools*.

cloudflared's bake is additionally a standing workaround: `Dockerfile:31-37` pins Cloudflare's **bookworm** apt suite on a trixie base because Cloudflare publishes no trixie suite (`pkg.cloudflare.com/cloudflared/dists/trixie` → HTTP 404). Moving it to a pinned, checksum-verified binary **deletes** that hack rather than carrying it.

## The blocker this has to solve first

#897's tailscale entry is the cited template, but copying it verbatim does not work, and the reason matters:

`tools/catalog.ts` root-installs to `/usr/local/bin` with `installUser: "root"`. Inside the sandbox that routes through `local-target.ts:113-116`, and because `commands/tool.ts:309` passes `stdio: "inherit"`, `argvFor` selects the **interactive** branch — plain `sudo --`, no `-n`. `/etc/sudoers.d/sandbox` grants `sandbox ALL=(ALL) ALL` with **no NOPASSWD**. Verified in a running sandbox:

```
$ sudo -n -- true
sudo: a password is required
```

So `oh tool install <any root tool>` from inside the sandbox hangs on a password prompt no agent can answer. This is the S1 risk carried out of #903 and it lands the moment a default tool is root-installed.

**This also affects #897's tailscale** (same shape, same `installUser: "root"`). Out of scope here; flagged for that PR.

## Approach

Install to `~/.local/bin` as the `sandbox` user, not `/usr/local/bin` as root — the same correction #900/#903 made for the harnesses. herdr and cloudflared are single static binaries with no system integration, so nothing requires a root-owned path. This:

- needs no sudo, so the operator-facing `oh tool install herdr` works from inside the sandbox;
- lands in the home mount, so an install survives container recreation instead of being re-downloaded on every boot;
- is upgradeable in place by a running sandbox, which a root-owned `/usr/local/bin` copy is not.

Changes:

- Add `kind: "default"` to `ToolKind` — provisioned at boot by the CLI; neither baked nor opt-in.
- herdr (0.7.4) and cloudflared (2026.8.2) become `kind: "default"`, `installUser: "sandbox"`, with per-arch pinned URLs and sha256 verification into `$NPM_USER_PREFIX/bin`. Checksums measured, not copied.
- Generalize #905's boot provisioner over both catalogs and rename it `provision-defaults.sh` (`OH_PROVISION_HARNESSES` → `OH_PROVISION_DEFAULTS`; neither has shipped, so no alias is owed).
- Delete the cloudflared apt block, the herdr `RUN`, and `ARG HERDR_VERSION` from the Dockerfile. The bookworm-suite workaround goes with them.
- Generalize the negative image oracle in `verify-sandbox-image.sh` and the boot-smoke assertion to cover default tools as well as default harnesses. The herdr version/sha pin moves from the Dockerfile to the catalog, and `herdr-default.test.ts` follows it.

`cc-safety-net` stays baked (operator ruling). `docker-cli` and `gh` stay `kind: "baked-in"` — both are apt packages with system integration, not single relocatable binaries.

## Cost this accepts

Worth stating plainly because it is worse than #905's: **an offline first boot on a fresh home mount now has no herdr**, and `herdr` is step 3 of the documented lifecycle. tmux is still present, so the sandbox is usable, but the documented first-run path needs network. The failure has to be a clear, actionable message rather than a broken `oh shell`.

ryaneggz:
Closes #906

**Stacked on #905** (`task/904-unbake-harnesses`, itself on #903). Retarget as the stack lands.

## Why

The operator ruling: *"outside the sandbox the CLI is used for provisioning host or sandbox on the host; once inside the sandbox the CLI is for provisioning sandbox with harnesses and tools."* herdr and cloudflared are tools. #905 did this for the default harnesses; this does it for the default tools.

cloudflared's bake was also a standing workaround — `Dockerfile:31-37` pinned Cloudflare's **bookworm** apt suite on a trixie base because Cloudflare publishes no trixie suite. Moving it to a pinned binary **deletes** that hack. Docker's is now the only third-party apt source in the image.

## The blocker, and why the cited template could not be copied

#897's tailscale entry was the plan's template. It does not work, and the reason is load-bearing:

`commands/tool.ts:309` installs with `stdio: "inherit"`, so `local-target.ts:113-116` picks the **interactive** branch — plain `sudo --`, no `-n`. `/etc/sudoers.d/sandbox` grants `sandbox ALL=(ALL) ALL` with **no NOPASSWD**. Verified in a live sandbox:

```
$ sudo -n -- true
sudo: a password is required
```

So `oh tool install <any root-installed tool>` hangs on a password prompt no agent can answer. This is the S1 risk carried out of #903, and it lands the moment a *default* tool is root-installed.

**#897's tailscale has the same defect** (`installUser: "root"`, same code path). Flagged for that PR; not fixed here.

The fix is the same correction #900 made for the harnesses: install to `~/.local/bin` as the `sandbox` user. herdr and cloudflared are single static binaries with no system integration, so nothing needs a root-owned path. This gets no sudo, survival across container recreation, and in-place upgradeability that a root-owned `/usr/local/bin` copy cannot have.

## What changed

- `ToolKind` gains `"default"`. herdr 0.7.4 and cloudflared 2026.8.2 become `kind: "default"`, `installUser: "sandbox"`, per-arch pinned URLs, `sha256sum -c -` into `$NPM_USER_PREFIX/bin`.
- `provision-harnesses.sh` generalizes over both catalogs → `provision-defaults.sh` (`OH_PROVISION_DEFAULTS`, timeout 180s → 240s). It `die`s rather than reporting success when neither catalog yields a default.
- Dockerfile loses the herdr `RUN`, `ARG HERDR_VERSION`, and the entire cloudflared apt block.
- Both oracles generalize: `verify-sandbox-image.sh` rejects a baked default harness **or tool**, reading each catalog out of the image; the boot smoke asserts every default in both catalogs resolves under `NPM_USER_PREFIX`, is owned by the sandbox uid, and prints a version.
- The herdr version + checksum pin moves from the Dockerfile to the catalog; `herdr-default.test.ts` follows it.

`cc-safety-net` stays baked (operator ruling). `docker-cli` and `gh` stay `baked-in` — apt packages with system integration, not relocatable single binaries.

## Checksums are measured, not asserted

Both binaries were downloaded and hashed, then both install scripts were **executed end to end** as the sandbox user against a throwaway prefix:

```
/tmp/tmp.kaKebhd5uy/herdr: OK          → herdr 0.7.4
/tmp/tmp.95hM6UnKO4/cloudflared: OK    → cloudflared version 2026.8.2
```

Negative test: corrupting the pinned sha gives `FAILED / WARNING: 1 computed checksum did NOT match`, exit 1, and **nothing installed**.

## Mutation-verified assertions

Six mutations against the new probe, each confirmed to exit 1 — plus one control confirmed to stay green:

| Mutation | Result |
|---|---|
| re-bake herdr in the Dockerfile | caught |
| re-bake cloudflared in the Dockerfile | caught |
| **control:** oh-my-zsh cloned from github.com | correctly still passes |
| default tool with `installUser: "root"` | caught |
| default tool without `sha256sum -c -` | caught |
| default tool installing to `/usr/local` | caught |

The control matters: my first draft matched on the bare host `https://github.com`, which the Dockerfile legitimately uses to clone zsh plugins. It failed against real content and I narrowed the match to the pinned project path.

## Cost this accepts

Worse than #905's, and worth stating plainly: **an offline first boot on a fresh home mount has no herdr**, and `herdr` is step 3 of the documented lifecycle. tmux remains, so the sandbox is usable. The entrypoint says so explicitly rather than failing silently:

```
[entrypoint] WARNING: default provisioning did not complete; run: bash .oh/scripts/provision-defaults.sh
[entrypoint] WARNING: herdr may be unavailable — 'tmux' still works as a fallback multiplexer
```

## Verification

`EVAL=0` (103 probes, no regressions). `npx vitest run` 953 passed / 8 failed — the known `compose-args.test.ts` environmental baseline. shellcheck 0, `tsc --noEmit` 0, `npm run build` 0. Local `docker build` and the booted-image checks follow in a comment.


ryaneggz:
## Verification results

### All five defaults provision at boot, on a fresh volume, in CI

```
Volume "openharness-boot-guard-33351597690_workspace"  Created
sandbox boot smoke: claude-code provisioned at boot -> 2.1.251 (Claude Code)
sandbox boot smoke: codex       provisioned at boot -> codex-cli 0.151.0
sandbox boot smoke: pi          provisioned at boot -> 0.84.4
sandbox boot smoke: herdr       provisioned at boot -> herdr 0.7.4
sandbox boot smoke: cloudflared provisioned at boot -> cloudflared version 2026.8.2 (built 2026-08-14-12:17 UTC)
sandbox boot smoke ok: ... boot-provisioned harness and tool checks
```

Both negative oracles fired in the same run, against the built image:

```
ok: no default harness is baked into the image (claude-code codex pi)
ok: no default tool is baked into the image (herdr cloudflared)
```

**Boot cost: 23 seconds** (02:45:29 → 02:45:52), against 21s for #905. Adding ~57 MB of binary downloads cost ~2s on a runner. 6/6 checks green.

### Local image

`docker build` succeeds; `verify-sandbox-image.sh` passes every check. `type -P` inside the image:

| Binary | Result |
|---|---|
| `claude`, `codex`, `pi`, `herdr`, `cloudflared` | absent |
| `cc-safety-net`, `gh`, `docker` | present |

`/etc/apt/sources.list.d/` now holds `debian.sources`, `docker.list`, `github-cli.list` — **`cloudflared.list` is gone**, which is the bookworm-suite workaround actually deleted rather than relocated.

### Size — a clean measurement this time

Unlike #905, both images here were built from the same checkout, so this is one-variable:

| | #905 image | this PR | Delta |
|---|---|---|---|
| `du -sx /` | 1,459,984 KB | 1,403,552 KB | **−55 MiB** |
| `/usr` | 930,704 KB | 873,200 KB | −56 MiB |
| `/opt`, `/root` | unchanged | unchanged | 0 |

The entire delta is in `/usr`, exactly where herdr (`/usr/local/bin`, 19 MB) and cloudflared (`/usr/bin`, 38 MB) lived. `/opt` and `/root` being byte-identical is the check that no build-context drift crept into this comparison — the flaw that made #905's top-line number unusable.

Docker's reported layer sum: 2.02 GB → 1.94 GB.

### Local checks

`EVAL=0` (103 probes, no regressions; `oh-config-surfaces` is the pre-existing persistent red). `npx vitest run` 953 passed / 8 failed — the known `compose-args.test.ts` baseline. shellcheck 0, `tsc --noEmit` 0, `npm run build` 0.

PR #909, opened 2026-08-31
FROM task/908-cli-first-class TO development

### LINKED ISSUE #908: task(#908): remove optional-harness build args from the image and make the CLI the install path
## Problem

#905 and #907 moved the `kind: "default"` harnesses and tools out of the image, but the four **optional** harnesses were left behind. The Dockerfile still carries `ARG INSTALL_HERMES`, `ARG INSTALL_DEEPAGENTS`, `ARG INSTALL_OPENCODE`, and `ARG INSTALL_GROK_BUILD` with a `RUN` block each. The operator's boundary — *"once inside the sandbox the CLI is for provisioning sandbox with harnesses and tools"* — has no carve-out for optional ones.

They are not merely leftover. Three of the four are `installUser: "root"`, and `harness.ts:256` installs with `stdio: "inherit"`, so `local-target.ts` selects the **interactive** branch. Evaluated against the real code:

```
user=root stdio="inherit" -> ["sudo","--","npm","install","-g","opencode-ai"]
$ sudo -n -- true
sudo: a password is required
```

`/etc/sudoers.d/sandbox` grants `sandbox ALL=(ALL) ALL` with no NOPASSWD, so `oh harness install opencode|grok-build|hermes` from inside the sandbox hangs on a password prompt no agent can answer. **The build arg is currently the only working path for those three**, which is why the Dockerfile blocks cannot simply be deleted.

Two smaller findings alongside:

- `buildArg` on `HarnessEntry` is **dead metadata** — declared on the interface, set on four entries, read by no code.
- `.oh/evals/probes/tool-catalog-boundary.sh:41` already *bans* `buildArg` in the tool catalog: *"that field carries a Dockerfile invariant this catalog cannot satisfy."* The two catalogs disagree on a rule one of them enforces.

## Approach

Every one of the four relocates to the sandbox user with no sudo. Verified by reading the upstream installers:

| Harness | Today | Relocation |
|---|---|---|
| opencode | `npm install -g` as root | `npm --prefix /home/sandbox/.local` — identical to claude-code |
| deepagents | already `installUser: "sandbox"` | already lands in `~/.local/bin` via `UV_TOOL_BIN_DIR`; its build arg is pure redundancy |
| grok-build | `HOME=/opt/grok-build GROK_BIN_DIR=/opt/grok-build/bin` | installer honours `GROK_BIN_DIR`; point it at `~/.local/bin` |
| hermes | root → `/usr/local/lib/hermes-agent` | installer honours `HERMES_INSTALL_DIR`, and its `get_command_link_dir()` already picks `$HOME/.local/bin` for a non-root install |

So no `NOPASSWD` change is needed and no security posture moves.

Changes:

- Delete all four `ARG`/`RUN` pairs from the Dockerfile, the matching entries from compose `build.args`, and the now-dead `/opt/grok-build` and `/usr/local/lib/hermes-agent` chowns in the final stage.
- Retarget all four catalog entries to `installUser: "sandbox"` under `$NPM_USER_PREFIX`, and remove the `buildArg` field from `HarnessEntry` entirely.
- Extend `provision-defaults.sh` to also install `kind: "optional"` harnesses whose `install.<key>` is true in oh.json, so declared intent is honoured on a fresh home mount. `isInstallFlagEnabled` already reads oh.json, and `oh harness list --json` already reports `enabled`, so this needs no new env plumbing.
- Generalize the image oracle from "no *default* harness is baked" to "no harness is baked at all".
- Retarget `sandbox-compatibility.yml`'s optional-installer job: it currently builds one image with every `--build-arg INSTALL_*=true` and asserts a version from each CLI. That subject disappears. Replace it with a booted container that runs `oh harness install <id>` for each optional and asserts the same versions — which tests the path operators actually use, instead of one that will no longer exist.

## Explicitly staying in the image

`INSTALL_HERMES` keeps its **runtime** life as a container environment variable — `link-providers.sh:111` vendors the Hermes skill pack from it and `entrypoint.sh:169` wires `auth.json`. Only its build-arg role goes away. `cc-safety-net`, `docker-cli`, `gh`, and the base toolchain stay baked; `INSTALL_PYTHON_KERNEL` is not a harness and is out of scope.

## Cost

An operator who has enabled an optional harness pays its download on a first boot against a fresh home mount, same as the defaults. Offline, they get the documented warn-and-continue.

ryaneggz:
Closes #908

**Stacked on #907.** Retarget as the stack lands.

## Why

#905 and #907 moved the `kind: "default"` harnesses and tools out of the image and left the four optional harnesses behind. Your boundary — *inside the sandbox the CLI provisions harnesses and tools* — has no carve-out for optional ones, and I had been scoping them out as though it did.

They were not merely leftover. **opencode, grok-build, and hermes are all `installUser: "root"`**, and `harness.ts:256` installs with `stdio: "inherit"`, so `local-target.ts` selects the *interactive* branch. Evaluated against the real code:

```
user=root stdio="inherit" -> ["sudo","--","npm","install","-g","opencode-ai"]
$ sudo -n -- true
sudo: a password is required
```

`/etc/sudoers.d/sandbox` has no NOPASSWD, so `oh harness install opencode` hangs on a password prompt no agent can answer. The build arg was the **only working path** — which is why the Dockerfile blocks could not simply be deleted, and why this PR is a fix as much as a cleanup.

## No sudoers change was needed

All four relocate to the sandbox user. I read the upstream installers rather than assuming:

| Harness | Was | Now |
|---|---|---|
| opencode | `npm install -g` as root | `npm --prefix /home/sandbox/.local` — identical to claude-code |
| deepagents | already `sandbox` + `uv tool install` | unchanged; its build arg was pure redundancy |
| grok-build | `HOME=/opt/grok-build GROK_BIN_DIR=…` | installer honours `GROK_BIN_DIR`; pointed at `~/.local/bin` |
| hermes | root → `/usr/local/lib/hermes-agent` | installer honours `HERMES_INSTALL_DIR`, and its own `get_command_link_dir()` already picks `$HOME/.local/bin` for a non-root install |

Every harness in the catalog is now `installUser: "sandbox"`. No security posture moves.

## What changed

- All four `ARG`/`RUN` pairs, the compose `build.args` block, and the now-dead `/opt/grok-build` and `/usr/local/lib/hermes-agent` chowns are gone.
- **`buildArg` removed from `HarnessEntry` entirely** — it was dead metadata: declared, set four times, read by no code. `tool-catalog-boundary.sh` already banned the same field in the tool catalog with *"that field carries a Dockerfile invariant this catalog cannot satisfy"*; the two catalogs no longer disagree.
- `provision-defaults.sh` reads the full catalog and installs any non-default entry whose `install.<key>` is true, so declared intent survives a fresh home mount. `isInstallFlagEnabled` already reads oh.json — no new env plumbing.
- `verify-sandbox-image.sh` widens to **"no harness of any kind is baked"**, and gains the inverse for tools: every `kind: "baked-in"` tool must be *present*, or the whole check would pass on an image missing everything.
- `sandbox-compatibility.yml`'s optional-installer job loses its subject. It now boots the image and runs `oh harness install <id>` for each optional harness, asserting the binary lands under `/home/sandbox/.local` — the path operators actually use, rather than one that no longer exists.

## Explicitly kept

`INSTALL_HERMES` keeps its **runtime** life as a container environment variable: `link-providers.sh:111` vendors the Hermes skill pack from it and `entrypoint.sh:169` wires `auth.json`. Only its build-arg role is gone, and a test now pins that distinction. `cc-safety-net`, `docker-cli`, `gh`, and the base toolchain stay baked. `INSTALL_PYTHON_KERNEL` is not a harness and is untouched.

## A test was enshrining the bug

`harness.test.ts` had a case named *"installs live instead of skipping the install"* asserting `calls.some((c) => c.cmd === "sudo")` — it codified the exact defect that made the command hang. It now asserts the opposite: no install shells out to sudo at all.

## Mutation-verified

Five against the widened provisioning probe, four against the CI-contract probe, each confirmed to exit 1:

| Mutation | Result |
|---|---|
| reinstate `ARG INSTALL_HERMES` | caught |
| re-bake `opencode-ai` in the Dockerfile | caught |
| optional harness back to `installUser: "root"` | caught |
| `buildArg` field returns | caught |
| hermes installs to `/usr/local` again | caught |
| compat job reintroduces a `--build-arg` | caught |
| compat job drops its vacuity guard | caught |
| compat job drops the home-mount location assertion | caught |
| compat job renamed away | caught |

The opt-in provisioning path was also exercised directly: setting `install.opencode: true` in oh.json made `provision-defaults.sh --verify` report opencode as missing, confirming declared intent is honoured rather than silently ignored.

## Verification

`EVAL=0` (103 probes, no regressions). `npx vitest run` 952 passed / 8 failed — the known `compose-args.test.ts` baseline. shellcheck 0, `tsc --noEmit` 0, `npm run build` 0. Local image build and booted-container checks follow in a comment.


ryaneggz:
## Verification results

### The image ships no harness at all

```
ok: no harness is baked into the image (claude-code codex pi opencode grok-build deepagents hermes t3code prime-agent)
ok: no default tool is baked into the image (herdr cloudflared)
ok: every baked-in tool is present (docker-cli gh)
verify-sandbox-image: all checks passed
```

The third line is the one that matters most — without it the first two would pass on an image missing everything.

### All four optional harnesses install through the CLI, into the home mount

Booted the built image and ran the exact flow the new CI job runs, as the `sandbox` user with no sudo anywhere:

| Harness | Resolved path | Owner | Version |
|---|---|---|---|
| opencode | `/home/sandbox/.local/bin/opencode` | sandbox | 1.18.25 |
| grok | `/home/sandbox/.local/bin/grok` | sandbox | grok 0.2.39 (55a20b703) |
| deepagents | `/home/sandbox/.local/bin/deepagents` | sandbox | deepagents-cli 0.3.0 |
| hermes | `/home/sandbox/.local/bin/hermes` | sandbox | Hermes Agent v0.20.6 (2026.8.27) |

Nothing leaked into a system path: `/usr/local/bin` holds none of them, and neither `/opt/grok-build` nor `/usr/local/lib/hermes-agent` exists.

**Hermes is the one I want to call out.** Its relocation was inferred from reading the upstream installer (`HERMES_INSTALL_DIR` plus `get_command_link_dir()` picking `$HOME/.local/bin` for non-root) rather than tested, so it was the highest-risk claim in the PR body. It installs and reports its version correctly.

### A near-miss in the CI job I wrote

`deepagents --version` prints an upstream `DeprecationWarning` as its **first** line and the actual version on its third. A version check that read only the first line would fail here. The job greps the whole captured output, so it matches — verified explicitly rather than assumed:

```
$ grep -Eq '(^|[^[:alnum:]])v?[0-9]+([.][0-9]+)+([^[:alnum:]]|$)' <<<"$out" && echo MATCH
MATCH
```

### This is not a size win, and I don't want to imply otherwise

| | #907 | this PR |
|---|---|---|
| `du -sx /` | 1,403,552 KB | 1,401,476 KB |
| layer sum | 1.94 GB | 1.93 GB |

~2 MB, which is noise. The build args all defaulted to `false`, so the default image was never carrying these harnesses — what this PR removes is the *ability* to bake them, not bytes that were being baked. The value here is ownership plus the sudo fix, not image size.

### Local checks

`EVAL=0` (103 probes, no regressions). `npx vitest run` 952 passed / 8 failed — the known `compose-args.test.ts` baseline. shellcheck 0, `tsc --noEmit` 0, `npm run build` 0.

### Separate finding, not fixed here

`deepagents-cli` is **deprecated upstream**: *"deepagents-cli is deprecated and will not receive further releases. Use the `managed-deepagents` package (`uv tool install managed-deepagents`, then `mda`) instead."* The catalog still points at the retired package. Out of scope for this PR; worth its own issue.

ryaneggz:
## Follow-up: the CI failure was in the job, not the change

The failed run was re-run **on the same SHA and passed**, verifying all four optional harnesses — hermes included, whose own installer logs `Hermes Agent itself does not require or retain root access.` So the relocation is sound; what I had built was a flaky required check.

`18da66d4` retries each install once before failing. I made a merge-blocking check depend on four third-party endpoints, and Hermes' installer hard-fails its entire run when its internal `npm install` blips (upstream does this deliberately — a silent partial install used to be worse). One retry absorbs that.

The contract is unchanged. A genuine break — wrong user, wrong path, a sudo prompt — fails both attempts and still fails the job. The probe asserts **both** halves, so neither the retry nor the hard failure after it can be dropped silently:

| Mutation | Result |
|---|---|
| retry loop reduced to one attempt | caught |
| retries but never fails hard | caught |

Now 6/6 green, `CLEAN` / `MERGEABLE`.

## Separate issue filed

[#910](https://github.com/mifunedev/openharness/issues/910) — `deepagents-cli` is deprecated upstream and directs users to `managed-deepagents` / `mda`. The catalog still installs the retired package. Left as an issue rather than fixed here: it is not a rename (the binary changes from `deepagents` to `mda`, moving `binary`, `verifyArgv`, the banner line, and the docs), and there is a prior question about whether the entry should follow upstream or be retired.

PR #938, opened 2026-09-01
FROM task/937-deployment-provisioning-guard TO development

### LINKED ISSUE #937: task: gate the published sandbox image on a live provisioning guard
## Problem

`ghcr.io/mifunedev/openharness:latest` is the artifact most users consume, and nothing
verifies it. Three gaps, confirmed at `4fbca95`:

1. **No CI job boots the image-only flavor.** Every live boot in CI goes through
   `.oh/scripts/docker-compose.sh`, which hardcodes `-f .devcontainer/docker-compose.yml`
   — a bind mount — so `entrypoint.sh:121` always takes the `if` branch. The `else`
   (seed) branch and `seed_workspace_volume` are exercised only by an in-process
   simulation inside `.oh/evals/probes/oh-image-only-deploy.sh`.
2. **No test anywhere runs the published image.** `sandbox-boot-guard.yml`,
   `sandbox-compatibility.yml`, and `release.yml` all verify a locally built tag. No
   workflow, script, or probe pulls a published GHCR ref.
3. `docs/deployment-prebuilt-image.md` ships a section literally titled
   **"Manual live-host smoke checklist (non-gating)"**. Per
   `.oh/knowledge/patterns/pattern-wiki-ungated-check-drift.md`, a report-only check that
   nothing gates on converges on never being run.

This is sharpened by `entrypoint.sh:176-181`: when `provision-defaults.sh` fails or times
out, boot logs a WARNING and continues. The container becomes healthy with the default
harnesses and tools missing, so healthiness alone is not evidence that provisioning worked.

A manual run of the blog recipe against v0.6.0 confirmed the path currently works, so this
is not a bug fix — it converts a one-off manual validation into a repeatable instrument and
closes the window in which a broken published image ships unnoticed.

## Scope

Resource provisioning (container, volume, network) and internal provisioning (defaults
installed inside the container). **Auth flows are explicitly out of scope.**

## Proposed shape

- `.oh/scripts/deployment-compose.sh` — image-only compose driver satisfying the existing
  `BOOT_SMOKE_COMPOSE` seam, pointed at the already-parameterised
  `.devcontainer/docker-compose.image-only.yml`.
- `.oh/scripts/sandbox-boot-smoke.sh` — one `BOOT_SMOKE_FLAVOR` switch; on `image-only`
  skip only `verify_bind_ownership`, which has no host checkout to compare against. Reuse
  `verify_default_catalog` as-is; it is already the exact provisioning oracle.
- `.oh/scripts/deployment-guard.sh` — operator entry point: resolve the image ref
  (`OH_SANDBOX_IMAGE`, default `:latest`), run `verify-sandbox-image.sh`, run the smoke in
  image-only flavor under a unique project name, then assert the seed branch, the
  `.image-seeded` marker, the container shape, and the `oh harness install` persist-and-install
  contract.
- `.oh/evals/probes/deployment-guard-ci.sh` — tier-A wiring guard (the live run cannot fit
  the 30s probe cap).
- `.github/workflows/deployment-guard.yml` — `workflow_run` after `Release` plus
  `workflow_dispatch`.
- `docs/deployment-prebuilt-image.md` — replace the non-gating checklist with a pointer to
  the guard.

## Non-goals

- Not a 10th `/audit` target — this executes a live readiness floor, which
  `.oh/knowledge/source/audit-architecture.md:33` keeps outside the audit taxonomy.
- No new scoreboard leg. The CI job is the record; git history is the time series.
- Auth flows.

ryaneggz:
Closes #937.

**Status: READY** — `/audit implementation` **AUDIT-PASS** (`audit-20260902T011610Z-2958187`, all 5 gates, `SIMPLICITY-RESIDUAL: 3`) · `/audit pr` **PR-AUDIT-PROMOTABLE** (`audit-20260902T012527Z-3041876`, classified against head `32b13bc7`) · `/eval` 140 probes rc=0, zero REGRESSION rows **and zero degradations** in a clean worktree · CI green (6/6), MERGEABLE, CLEAN.

> ⚠️ **`/benchmark` says NOT-BENEFICIAL.** The regression floor is clear, but the capability-benchmark ceiling is flat and **no capability task credits this change** — CB-001 scores unattended autopilot, CB-002 the `/spec` pipeline, CB-005 the retro→pattern→probe loop; none covers deployment readiness. Per the rubric that is *"machinery without benchmark movement"*. The honest reading is a **gap in the instrument** — but that is the `CB-004` failure mode this scoreboard already records, where machinery stood at `Δ +0.00` for two months because no measurement was taken. **A deployment-readiness capability task is the named follow-on.** `/benchmark` is read-only; the revert is the operator's call.

## What this is

A **parent-sandbox QA door**: a parent harness stands up a child sandbox over the Docker socket, asserts against it, and destroys it, so QA never pollutes the parent. `/deploy-check` is the door; `.oh/scripts/deployment-guard.sh` is the mechanism; there is **no CI leg**, by design.

Before this, nothing verified `ghcr.io/mifunedev/openharness:latest` — the artifact the docs tell users to pull. No CI job booted the image-only flavor (every live boot routes through the bind-mount compose file, so `entrypoint.sh` always took the *checkout bind detected* branch), nothing anywhere pulled a published ref, and the deployment doc shipped a checklist headed *"non-gating"*.

## What the door found on its first real use

`oh tool install agent-browser` failed in a fresh sandbox with:

```
ERROR  EACCES: permission denied, mkdir '/usr/local/share/pnpm'
```

`agent-browser` is the only catalog entry installed with `pnpm add -g`, and pnpm honours `PNPM_HOME`, not `NPM_USER_PREFIX`. `PNPM_HOME` was `/usr/local/share/pnpm` — root-owned, and absent from the image — while the catalog sets `installUser: "sandbox"`. **It could not succeed for any non-root user on any current image.** The parent sandbox is the one machine that cannot reproduce it: its copy dates to 2026-07-09, baked before harnesses were unbaked.

Fixed by moving `PNPM_HOME` into the home mount, sandbox-owned. Verified in a child from a locally built image: install exits 0, `agent-browser 0.8.5`, catalog `installed: true`.

Second half: the failure was invisible where it mattered. `entrypoint.sh` downgrades a provisioning failure to a WARNING so a transient outage cannot abort boot — right call — but the container then reported `(healthy)` with a requested tool absent. A marker now fails the healthcheck and names the recovery. Observed live: `(unhealthy)`, then `provision-defaults.sh` clears it.

**Behavior change worth knowing:** the boot smoke and `oh sandbox` wait for healthy, so a genuine registry outage now surfaces as a boot that never goes healthy rather than one that quietly comes up short.

## And a defect in the audit's own gate

Gate 4 failed with `FAIL gate4: Chromium launch` — **the gate could not pass for any unit**. Two defects: the isolated `HOME` hid Playwright's browser cache, and the `agent-browser` daemon socket under `AUDIT_TMP_ROOT` overflowed the 107-byte unix path limit. Both fixed. The existing probe passed throughout because its mocked `agent-browser` exits 0 — `pattern-evals-unexercised-oracle`, in the subsystem that enforces it. The probe now asserts the environment the preflight hands the browser; 4 fault injections, 4 caught.

## Evidence

| Check | Result |
|---|---|
| Guard vs `:latest` | 32 `ok:` lines, exit 0 |
| Guard vs `:0.5.1` (negative control) | exit 1, 6 named failures |
| Interrupt / collision tests | daemon byte-identical, nothing leaked |
| Guard vs a clean-tree local build (`--local`) | all checks passed, exit 0, no pull |
| Guard vs a dirty-tree local build (`--local`) | exit 1, correctly naming the un-provisioned defaults |
| `deployment-guard-ci.sh` injections | 22/22 caught, baseline green |
| `provision-marker-fail-closed.sh` injections | 2/2 caught, baseline green |
| `audit-implementation-behavior.sh` injections | 4/4 caught, baseline green |
| Gate 5 simplify rounds | round 1 `netAdded` 1830 → 1711 (comments 98 → 20); round 2 stopped by the monotone rule |

- `.oh/tasks/deployment-provisioning-guard/evidence.md` — per-gate observed output, the divergences, and what remains unverified
- `.oh/tasks/deployment-provisioning-guard/progress.txt` — the per-story narrative

## Where it diverged from the plan

Three operator-directed changes: `/deploy-check` was added as the primary door; the CI leg was then removed entirely; and the repair of what the door found was directed here rather than into a separate change. Plus five implementation deviations recorded in `evidence.md` — the boot smoke's `trap` destroying the container the guard needed, narrow `docker inspect` templates after a guard rejected a full dump, asserting persistence through the CLI's own `enabled` field, a doc line that was incomplete rather than wrong, and `${OH_DEFAULT_SANDBOX_IMAGE:-…}` after a probe caught a hardcoded registry.

## Local image validation, and the two defects it surfaced

`--local` now closes the gap this PR originally shipped as documented-but-unfixed. The guard takes an image already on the daemon, asserts it is present, and reports its id and build time. It deliberately does **not** fall back: a failed pull never silently resolves to whatever is on the daemon, because that validates a stale object and reports it as the published image.

The first live `--local` run went red — and was right to. The Dockerfile does `COPY . /opt/oh-seed/`, so an image built from a dirty checkout bakes the working tree; the uncommitted `oh.json` (`hermes=true`) rode in, the child opted into hermes, whose installer prompts on a tty, and it consumed the 240s provisioning budget before `herdr` and `cloudflared` were reached. An artifact of the build, not the recipe — now a documented `--local` caveat. The confirming run from a clean worktree passes every assertion at exit 0.

Watching that failure surfaced something that **was** a defect, and a hole in this PR's own healthcheck fix: the provision-failed marker was written only at the *end* of `provision-defaults.sh`, lines a SIGTERM never reaches. Boot provisioning runs under `timeout`, so a kill part-way through an install is the likeliest failure — and the container reported `sandbox healthcheck ok` with defaults missing. The marker is now fail-closed: written before the first install, removed only after a clean completion.

`/audit implementation` also caught a practice worth naming: the committed `RESULTS.md` had one row hand-corrected for an environment-caused red. It was right — a hand-corrected scoreboard is not a run record. The dilemma behind it was false; the four `PASS→SKIPPED` degradations came from an *uninstalled* worktree, not a clean one. With dependencies installed and the CLI built, the suite runs 140 probes at exit 0 with zero regressions and zero degradations, and `RESULTS.md` is that run verbatim.

## What remains unverified

- **No "leave me a child to poke at" mode.** Every reproduction in this branch was driven by hand through `deployment-compose.sh`.
- **Only `provisioning` exists.** The scenario argument is the documented extension point but has one value, so extensibility is design, not demonstration.
- **One blocking installer starves every later catalog entry.** Real and independent of the dirty `oh.json` — any default could hang the same way. A per-install `timeout` would bound it; that changes provisioning policy for every sandbox and is not needed to close the `--local` gap.
- **The hermes installer prompts on a tty.** Upstream behavior reached through a `curl | bash` installer; opt-in, and not this PR's subject.
- **Gate 5 residual 1 declined on measurement.** `compose config --format json` emits Go duration *strings* (`10m0s`), so replacing the `awk` healthcheck scrape needs a duration parser — more branch points, not fewer.
- **Gate 5 residual 2 declined on design.** Folding the image-only driver into `docker-compose.sh` would put an image-only mode inside the operator's own lifecycle door, and the dotenv divergence US-001 requires is exactly what that door exists to do.
- **The gate 4 applicability grep was left alone.** It false-positives on `agent-browser`; narrowing it risks silently skipping real UI verification.
- **`compose-config-path-parity` is red in the authoring sandbox only** — an untracked machine-local root dotenv makes `docker-compose.sh:62-63` prefer it while VS Code path B loads only `.devcontainer/.env`, so the two paths really do diverge there. A true report about that machine. PASS in a clean worktree of this commit and in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via /spec execute


ryaneggz:
Returned to draft: three commits landed after the promotable classification, so the verdict no longer describes this head.

- `7c1ba1fc` removes the deployment-guard CI leg (operator direction: *"lets simplify without ci for now"*). The instrument stays a parent-sandbox QA door.
- `5f202c2d` fixes the defect that door found on its first real use: `PNPM_HOME` was a root-owned system path, so `oh tool install agent-browser` died with `EACCES: permission denied, mkdir '/usr/local/share/pnpm'` for every non-root user on every current image. Also stops a failed provision reporting `(healthy)`.
- `86057a5b` reconciles the PRD, task graph, skill, and docs with both.

Re-entering the promotable gate: CI on the new head, then a fresh `/audit implementation` and `/audit pr`. Note `sandbox-boot-guard.yml` now builds and boots the changed Dockerfile, which is the real check on the `PNPM_HOME` move.

ryaneggz:
Re-audited on `c6943b68` — `/audit implementation` run `audit-20260901T190538Z-1133824`.

**Gates 1, 2, 3 PASS.** 10/10 stories; 139 probes with zero REGRESSION rows in a clean worktree (including `compose-config-path-parity`, confirming that red is authoring-checkout-only); `promotable: true`, `evidenceComplete: true`, CI PASS, MERGEABLE, CLEAN.

**Gate 5 PASS with `SIMPLICITY-RESIDUAL: 1`.** The previous round's three findings were applied: `netAdded` 1830 → 1711, `deployment-guard.sh` 363 → 307 lines, comments 56 → 4. The residual is `shBranchPoints +101` with no concrete smaller alternative named.

**Gate 4 FAIL — and it is the gate that is broken, not this unit.** Verified independently in this checkout:

1. **The applicability oracle false-positives on this task's own subject.** `implementation-gates.sh:45` is `grep -qi 'agent-browser\|Verify in browser'` over `prd.json`. US-010 is about *installing the `agent-browser` CLI*; no story declares a browser-verification criterion and there is no application to navigate.
2. **The preflight cannot pass in this environment for any unit.** `implementation-gates.sh:72-76` runs `HOME="$profile" agent-browser open about:blank` against a fresh empty profile, while Playwright resolves its browser cache at `$HOME/.cache/ms-playwright` — which exists only under the real `HOME` (`chromium_headless_shell-{1200,1228,1234}` are present there). No `PLAYWRIGHT_BROWSERS_PATH` is set anywhere in the script, so the isolation contract and the cache lookup are mutually unsatisfiable.

`agent-browser --version` → `0.8.5`, and the same `open about:blank` succeeds under the real `HOME`.

Remediation belongs in `implementation-gates.sh browser-preflight` (seed `PLAYWRIGHT_BROWSERS_PATH` at the shared cache while keeping the profile isolated), not in this diff. Deliberately **not** rewording US-010 to dodge the grep — that would game the gate rather than fix it.

Staying draft until that is resolved.

Issue #956, opened 2026-09-04
feat: make systemd PID 1 and move cron supervision out of tmux


ryaneggz:
## Summary

Make the default OpenHarness Docker sandbox a **systemd-based OS container**:

- `systemd` becomes PID 1;
- the existing OpenHarness bootstrap runs as a systemd oneshot;
- `cron-runtime.ts` runs directly under a native systemd service;
- the `cron-watchdog` tmux supervisor is retired;
- the scheduler-level `cron-system` tmux session is retired;
- existing per-fire `tmux: true` cron behavior remains unchanged.

This issue is intentionally scoped to **container init + infrastructure supervision**. It does not add Tailscale behavior, convert OpenHarness cron definitions into systemd timers, or redesign other runtime substrates.

The architectural rule this establishes is:

> **systemd owns OS/container process supervision; OpenHarness owns scheduling semantics. tmux remains for interactive/detached user-agent sessions, not infrastructure watchdogs.**

---

## Current state

The Docker sandbox currently boots through:

```text
Docker
  ↓
Tini (`init: true`)
  ↓
entrypoint.sh
  ↓
sleep infinity
```

Both `.devcontainer/docker-compose.yml` and `.devcontainer/docker-compose.image-only.yml` explicitly set:

```yaml
init: true
entrypoint: /usr/local/bin/entrypoint.sh
command: sleep infinity
```

The entrypoint also owns cron process supervision:

```text
entrypoint.sh
  ↓
tmux cron-watchdog
  ↓
tmux cron-system
  ↓
node --experimental-strip-types .oh/scripts/cron-runtime.ts
```

The watchdog exists only to detect a missing `cron-system` tmux session and restart it. Once systemd is PID 1, that is duplicate supervision machinery.

The scheduler itself already has the process semantics needed for systemd supervision:

- exclusive PID locking via `crons/.pid`;
- `SIGTERM` / `SIGINT` cleanup;
- `SIGHUP` reload;
- dynamic Markdown cron loading;
- overlap policy;
- preflight handling;
- provider selection/fallback;
- per-fire tmux behavior when `tmux: true`.

So the target is to replace the **supervision layer**, not the cron domain model.

---

## Target architecture

```text
Docker
  ↓
systemd                                   PID 1
  ├─ openharness-bootstrap.service        oneshot
  │    └─ entrypoint.sh
  │         ├─ seed home
  │         ├─ UID/GID reconciliation
  │         ├─ provider-link/bootstrap work
[...]

PR #976, opened 2026-09-06
FROM bug/975-docker-socket-gid-alignment TO development

### LINKED ISSUE #975: bug(975): entrypoint docker socket GID alignment fails silently when the GID is already taken
## Symptom

Inside a sandbox with `access.dockerSocket: true`, the `sandbox` user cannot reach `/var/run/docker.sock` without `sudo`:

```
$ docker ps
permission denied while trying to connect to the docker API at unix:///var/run/docker.sock
$ sudo docker ps   # works
```

## Cause

`.devcontainer/entrypoint.sh:4-11` aligns the container `docker` group to the socket's GID:

```sh
groupmod -g "$HOST_GID" docker 2>/dev/null || true
```

`groupmod -g` refuses a GID that another group already holds, and `2>/dev/null || true` swallows the failure. The sandbox user stays in `docker` (GID 1001) while the socket needs GID 999.

The collision is deterministic, not incidental. In `.devcontainer/Dockerfile`, `apt-get install systemd systemd-sysv dbus` (:10-18) and `cron` allocate system GIDs downward from 999 (`systemd-journal:999`, `systemd-network:998`, `crontab:997`) *before* `groupadd -f docker` (:58-62) runs with no `-g` and lands on 1001. So 999 — the most common host docker GID on Debian/Ubuntu — is guaranteed to be taken by `systemd-journal` before the `docker` group exists.

## Impact

`/etc/sudoers.d/sandbox` is `sandbox ALL=(ALL) ALL`, deliberately **not** NOPASSWD (probe-enforced in `harness-one-door.sh:92` and `tailscale-tool-boundary.sh:11,137`). A non-interactive agent hits the password prompt and fails, so this is not "use sudo instead" — Docker is unreachable for automation, which is the case the socket mount exists to serve.

Observed live: socket `root:systemd-journal` (999), `id sandbox` → `1000(sandbox),1001(docker)`, `docker info` as `sandbox` fails.

## Fix

`.devcontainer/entrypoint.sh:137-162` already solves the identical problem correctly for UID/GID sync: probe `getent group "$GID"` first, join the incumbent group when the GID is claimed, renumber only when free, and route every mutation through `uid_reconcile_step` so failures print `[entrypoint] WARNING:` instead of vanishing. Apply that same pattern to the docker socket block.

Nothing covers `entrypoint.sh:4-11` today — no probe, no test, no documentation.

ryaneggz:
Closes #975

## Summary

The sandbox user could not reach a mounted Docker socket without `sudo`. `.devcontainer/entrypoint.sh` aligned the container `docker` group to the socket's GID with `groupmod -g "$HOST_GID" docker 2>/dev/null || true`. `groupmod -g` refuses a GID that another group already holds, and `2>/dev/null || true` swallowed the failure, leaving `sandbox` in a `docker` group that did not own the socket.

The collision is deterministic, not incidental. In `.devcontainer/Dockerfile`, `apt-get install systemd systemd-sysv dbus` (`:10-18`) and `cron` allocate system GIDs downward from 999 (`systemd-journal:999`, `systemd-network:998`, `crontab:997`) **before** `groupadd -f docker` (`:58-62`) runs with no `-g` and lands on 1001. So 999 — the most common host Docker GID on Debian/Ubuntu — is guaranteed taken before the `docker` group exists. Making systemd PID 1 (#956) is what turned this from likely into certain.

Impact is worse than a missing convenience: `/etc/sudoers.d/sandbox` is `sandbox ALL=(ALL) ALL`, deliberately **not** NOPASSWD (probe-enforced in `harness-one-door.sh:92` and `tailscale-tool-boundary.sh:11,137`). A non-interactive agent hits the password prompt and fails, so Docker is unreachable for exactly the automation the socket mount exists to serve.

The fix mirrors the UID/GID reconciliation already in this file at `:137-162`: probe `getent group "$SOCK_GID"` first, join the incumbent group when the GID is claimed, renumber `docker` only when it is free, and route both mutations through `uid_reconcile_step` so a failure prints `[entrypoint] WARNING:` instead of vanishing.

### Approaches considered

`groupmod -o -g` (duplicate GID) was rejected: `id` and `ls -l` resolve a GID through `getgrgid()`, which returns the first `/etc/group` entry with that GID, and `systemd-journal` is written at image build long before `docker` — so `id sandbox` prints `999(systemd-journal)` either way. It buys a duplicate-GID `/etc/group` and returns nothing legible. Mutating the socket's group or ACL was rejected outright: the socket is a bind mount, so that writes to the **host** inode. A build-time GID pin does not help, because the host GID is unknowable at build time.

All candidates are privilege-equivalent — the kernel compares numeric GIDs and never group names — so the choice turns on robustness, legibility, and consistency with the existing pattern, not on security.

### `set -e` note for reviewers

`entrypoint.sh:2` sets `-e`, so both calls sit in `if` conditions. Writing `uid_reconcile_step … && echo …` instead would abort the boot on failure, with no retry: the unit is `Type=oneshot` and never re-runs. That would turn a silent-degradation bug into a container that will not start.

## Verification

- `pnpm run test:scripts` — 938 passed / 60 files, including a new assertion in `.oh/scripts/__tests__/entrypoint.test.ts`.
- `pnpm run typecheck` — clean.
- `shellcheck -S warning .devcontainer/entrypoint.sh` — clean (v0.11.0), matching the `boot-lint` job.
- New probe `.oh/evals/probes/docker-socket-gid-alignment.sh` is a hybrid: the structural half asserts unconditionally and can return 1, so it is never an always-SKIP row in CI; the behavioural half compares `id -G sandbox` against the live socket's GID when a socket is mounted.
- Fault injection, each driven red against the committed change and restored:
  1. original swallowing `groupmod` → names the missing incumbent-group branch, the silenced stderr, and the `|| true`
  2. helper un-hoisted below its call site → `uid_reconcile_step is defined at line 22, below the docker socket block at line 4`
  3. `usermod -aG` → `-G` → `replaces sandbox's supplementary groups instead of appending`
  4. vitest assertion deleted → `lost the docker socket GID alignment assertion`
- Live behavioural check in a sandbox with `access.dockerSocket: true`: socket `root:systemd-journal` (999), `id sandbox` → `1000 1001`, probe REGRESSION and `docker info` fails as `sandbox`. This is the bug reproducing against real state.

Boot-path behaviour is proven by `sandbox-boot-guard`, not by a local Docker run.

## Out-of-scope change, called out

`CHANGELOG.md` — the `[Unreleased]` entry from #974 is 488 characters against the 250-character cap that `changelog-entry-length.sh` enforces, so `eval-probes` was already red on `development` and no PR touching `.oh/**` or `.devcontainer/**` could be green. This PR trims that entry to restore the gate. No other content changed.

## Docs

- `.env.example` — the claim that the sandbox user "is also in the `docker` group" becomes false whenever the host GID is taken; corrected to "the group that owns the socket". This is the only place documenting that escalation path.
- `docs/security-considerations.md` §4 Caveat 1 — records that alignment is by numeric GID, so on Debian images `sandbox` joins `systemd-journal` and incidentally gains journal read. That grant is inherent to any correct fix and was previously undocumented.

## Checklist

- [x] Base branch is `development`
- [x] Title is `FROM <source-branch> TO <target-branch>`
- [x] Closing keyword links every issue this PR completes
- [x] `CHANGELOG.md` has an `## [Unreleased]` entry

https://claude.ai/code/session_018hyGq8qEWaHV4u1dpSagHp


ryaneggz:
## Behavioural verification on live state

The probe's behavioural half is now proven end to end against a real mounted socket, closing the one gap the original PR body listed as outstanding.

**Before** — the bug reproducing, with the container booted from the unfixed entrypoint:

```
$ ls -l /var/run/docker.sock
srw-rw---- 1 root systemd-journal 0 /var/run/docker.sock
$ id sandbox
uid=1000(sandbox) gid=1000(sandbox) groups=1000(sandbox),1001(docker)
$ docker ps
permission denied while trying to connect to the docker API at unix:///var/run/docker.sock

$ bash .oh/evals/probes/docker-socket-gid-alignment.sh
REGRESSION: /var/run/docker.sock is owned by GID 999 (systemd-journal) but sandbox's groups are
1000 1001 — docker is unreachable for a non-interactive agent
```

**After** — applying by hand exactly what the fixed entrypoint does at boot (`usermod -aG systemd-journal sandbox`), then a fresh process so the supplementary groups are picked up:

```
$ id
uid=1000(sandbox) gid=1000(sandbox) groups=1000(sandbox),999(systemd-journal),1001(docker)

$ bash .oh/evals/probes/docker-socket-gid-alignment.sh
PASS: sandbox holds GID 999, the group that owns /var/run/docker.sock, and the entrypoint
reconciles it without swallowing failures

$ docker ps
CONTAINER ID   IMAGE                                  COMMAND                  ...
502d16224fbe   ghcr.io/mifunedev/openharness:latest   "/usr/local/bin/entr…"   ...
[...]
```

`docker ps` runs as uid 1000 with no `sudo`, which is the outcome the change exists to produce.

## Unrelated observation from the same container

That sandbox reports `(unhealthy)`, and it is not caused by this branch. Its PID 1 is `docker-init`, so it predates #956 making systemd PID 1 (or was started through a path that bypasses `/sbin/init`), while `sandbox-healthcheck.sh` asserts `openharness-bootstrap.service` and `openharness-cron.service` are active units:

```
System has not been booted with systemd as init system (PID 1). Can't operate.
sandbox healthcheck failed:
- systemd unit not active: openharness-bootstrap.service
- systemd unit not active: openharness-cron.service
```

It read `(healthy)` earlier only because it was inside the healthcheck's start period. Recreating the sandbox resolves it, and also makes the alignment above happen at boot rather than by hand.

